PR #165, opened 2023-01-17
Accept and verify JSON web tokens

mtbc:
Resolves CHRON-204, CHRON-205, CHRON-206, and CHRON-209.

Introduces feature `devmode` that implies both `inmem` and new feature `anonymous-api` which allows unauthenticated access to the GraphQL API by default. Without selecting that feature, the `--anonymous-api` option to `serve-graphql` is required to allow it.

Otherwise, access to the API must now be authenticated: supply instead the `--jwks-address` option with a URI to the keyset used to sign the JWT bearer tokens that one's providing via `Authorization:`. The `--id-pointer` option must be used to provide a JSON pointer into the JWT claims for the Chronicle identity. An `AuthId` bearing the verified external ID can then be found in the GraphQL context.

The Apollo Studio sandbox from `--open` is removed but the Altair GraphQL client is a fine substitute that makes it easy to pass extra header lines.

### DIFF
--- a/crates/api/src/chronicle_graphql/authorization.rs
+++ b/crates/api/src/chronicle_graphql/authorization.rs
@@ -23,8 +23,8 @@ pub enum Error {
         #[from]
         source: jwtk::Error,
     },
-    #[error("application error: {0}", message)]
-    Simple { message: String },
+    #[error("formatting error: {0}", message)]
+    Format { message: String },
 }
 
 pub struct JwtChecker {
@@ -38,7 +38,7 @@ impl JwtChecker {
         }
     }
 
-    #[instrument(skip(self))]
+    #[instrument(skip(self), ret(Debug))]
     pub async fn verify_jwt(
         &self,
         token: &str,
@@ -51,7 +51,7 @@ impl JwtChecker {
             .map(|component| base64_engine.decode(component))
             .collect::<Result<Vec<Vec<u8>>, base64::DecodeError>>()?;
         if components.len() != 3 {
-            return Err(Error::Simple {
+            return Err(Error::Format {
                 message: format!("JWT has unexpected format: {}", token),
             });
         };
@@ -63,7 +63,7 @@ impl JwtChecker {
         if let Value::Object(claims) = serde_json::from_slice(components[1].as_slice())? {
             Ok(claims)
         } else {
-            Err(Error::Simple {
+            Err(Error::Format {
                 message: format!("JWT claims have unexpected format: {:?}", components[1]),
             })
         }
--- a/crates/api/src/chronicle_graphql/authorization.rs
+++ b/crates/api/src/chronicle_graphql/authorization.rs
@@ -1,0 +1,71 @@
+use base64::Engine;
+use jwtk::jwk::RemoteJwksVerifier;
+use serde_json::Value;
+use std::time::Duration;
+use thiserror::Error;
+use tracing::instrument;
+use url::Url;
+
+#[derive(Debug, Error)]
+pub enum Error {
+    #[error("Base64 decoding failure: {0}", source)]
+    Base64 {
+        #[from]
+        source: base64::DecodeError,
+    },
+    #[error("JSON decoding failure: {0}", source)]
+    Json {
+        #[from]
+        source: serde_json::Error,
+    },
+    #[error("JWT validation failure: {0}", source)]
+    Jwks {
+        #[from]
+        source: jwtk::Error,
+    },
+    #[error("application error: {0}", message)]
+    Simple { message: String },
+}
+
+pub struct JwtChecker {
+    verifier: RemoteJwksVerifier,
+}
+
+impl JwtChecker {
+    pub fn new(jwks_uri: &Url) -> Self {
+        Self {
+            verifier: RemoteJwksVerifier::new(jwks_uri.to_string(), None, Duration::from_secs(100)),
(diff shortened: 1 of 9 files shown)


crates/api/src/chronicle_graphql/authorization.rs
```diff
@@ -0,0 +1,68 @@
+use base64::Engine;
+use jwtk::jwk::RemoteJwksVerifier;
+use serde_json::Value;
+use std::time::Duration;
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+pub enum Error {
+    #[error("Base64 decoding failure")]
```
ryanbtp, line 9: Print the inner error here with {0}, it's a little infuriating you panic or have otherwise limited error reporting without the inner in the message otherwise

crates/api/src/chronicle_graphql/authorization.rs
```diff
@@ -0,0 +1,68 @@
...
+    #[error("application error")]
+    Simple { message: String },
+}
+
+pub struct JwtChecker {
+    verifier: RemoteJwksVerifier,
+}
+
+impl JwtChecker {
+    pub fn new(jwks_uri: &str) -> Self {
```
ryanbtp, line 33: Bring in url:Url and use that so we can validate further up the stack

crates/api/src/chronicle_graphql/authorization.rs
```diff
@@ -0,0 +1,68 @@
...
+    ) -> Result<serde_json::map::Map<String, Value>, Error> {
+        let base64_engine = base64::engine::general_purpose::URL_SAFE_NO_PAD;
+
+        // JWT is composed of three base64-encoded components
+        let components = token
+            .split('.')
+            .map(|component| base64_engine.decode(component))
+            .collect::<Result<Vec<Vec<u8>>, base64::DecodeError>>()?;
+        if components.len() != 3 {
+            return Err(Error::Simple {
```
ryanbtp, line 51: Probably worth renaming this to Format rather than Simple?

crates/api/src/chronicle_graphql/mod.rs
```diff
@@ -366,6 +377,49 @@ async fn gql_playground() -> impl IntoResponse {
...
+where
+    Q: ObjectType + 'static,
+    M: ObjectType + 'static,
+    S: SubscriptionType + 'static,
+{
+    type Output = GraphQLBatchResponse;
+
+    async fn call(&self, req: poem::Request) -> poem::Result<Self::Output> {
+        if let Some(authorization) = req.header("Authorization") {
+            let authorization = HeaderValue::from_str(authorization).unwrap();
```
ryanbtp, line 399: No unwraps in this sort of code please, that's an easy DOS attack

ryanbtp: Can we span / instrument this section so it's easy to get timings out of it. Either use the span!().in_scope(|| or extract to another fn and #[insurement] it

ryanbtp: This refers to the verify_jwt bit, we should get decent enough spans once the request is executng

crates/api/src/chronicle_graphql/authorization.rs
mtbc: I was thinking of it as maybe a catchall for any other application-initiated errors that come up.

crates/api/src/chronicle_graphql/authorization.rs
mtbc: Thanks, still learning how to use `thiserror`!

mtbc:
Feel free to message me if you'd like a JWT and a JWKS URI for testing.

crates/api/src/chronicle_graphql/authorization.rs
ryanbtp: Name them, it's much better to explicitly have separate types for errors or it ends up with the Go / rust-sawtooth-sdk pattern of error basically being String. Rust can be much more useful. It's also nice to be able to look at a modules Error type and see what kind of Bad Things can potentially happen. Error("Bad thing") doesn't have the same utility

ryanbtp: Current case in point - sawtooth sdk + zmq sending me UknownError("Unknown") which is plainly the result of this sort of thing somewhere (or possibly poor c interop)

crates/api/src/chronicle_graphql/authorization.rs
```diff
@@ -0,0 +1,71 @@
...
+}
+
+impl JwtChecker {
+    pub fn new(jwks_uri: &Url) -> Self {
+        Self {
+            verifier: RemoteJwksVerifier::new(jwks_uri.to_string(), None, Duration::from_secs(100)),
+        }
+    }
+
+    #[instrument(skip(self))]
```
ryanbtp, line 41: adding ret(Debug) here should eliminate the need to explictly debug!() it after, I also make this mistake a lot

mtbc:
Lost a commit there for a moment!

ryanbtp (approved):

crates/api/src/chronicle_graphql/mod.rs
mtbc: Thanks, was worth the rewrite to eliminate `unwrap()`.

PR #168, opened 2023-01-19
refactor: move identity to its own module

suchapalaver:
Cleans up `common` by moving all `AuthId` code to a separate identity.rs module.

### DIFF
--- a/crates/common/src/lib.rs
+++ b/crates/common/src/lib.rs
@@ -8,6 +8,7 @@ pub mod attributes;
 pub mod commands;
 pub mod context;
 pub mod database;
+pub mod identity;
 pub mod ledger;
 pub mod protocol;
 pub mod prov;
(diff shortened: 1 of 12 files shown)


crates/common/src/lib.rs
```diff
@@ -1,4 +1,4 @@
-#![cfg_attr(feature = "strict", deny(warnings))]
+// #![cfg_attr(feature = "strict", deny(warnings))]
```
mtbc, line 1: Do we want to leave this here as a comment?

mtbc:
Looks good overall, thank you.

crates/common/src/lib.rs
suchapalaver: Nice catch! Thanks! Fixed.

ryanbtp (approved):

PR #169, opened 2023-01-19
wip: anonymous identity variant

suchapalaver:


mtbc:
:+1: for the plan to hide use of anonymous identities behind `devmode` feature.

mtbc:
So, when acting as an anonymous entity, a signing key may possibly be supplied separately by the user/caller rather than being determined by (looking up something based on) the `AuthId` or just made up on demand?

PR #175, opened 2023-01-25
adjust how suffixing is done in the model

mtbc:
In reading domain definitions, no longer suffixes `external_id` with `Agent`, `Entity`, `Activity`, instead suffixes them by need. This prevents excessive suffixing as one translates repeatedly between Chronicle's internal representation and that of `domain.yaml` files. It also solves CHRON-210's issue with CLI help text.


suchapalaver (approved):
Nice!

PR #183, opened 2023-01-28
perf: rewrite opa-executor to bootstrap a built opa instance

suchapalaver:
Rewrites, refactoring in places, Chronicle's `OpaExecutor` and `PolicyBuilder` implementations with a view to bootstrapping a built `Opa` instance.

Adds benchmarks for `OpaExecutor` and `PolicyBuilder` methods. Run `cargo bench` for output.

[CHRON-223](https://blockchaintp.atlassian.net/browse/CHRON-223)

[CHRON-223]: https://blockchaintp.atlassian.net/browse/CHRON-223?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mtbc:
When `cfg!(feature = "devmode") && !options.is_present("opa-rule")` might be good to log warn of insecure operating mode.

mtbc:
It may be nice to `#[instrument]` the `build_opa` method given that it might take even rather longer for a non-trivial policy.

suchapalaver:
Thanks for those suggestions, @mtbc!

mtbc:
Thank you for effecting them. :smile:

crates/chronicle/src/bootstrap/mod.rs
```diff
@@ -212,22 +229,15 @@ pub async fn graphql_server<Query, Mutation>(
     options: &ArgMatches,
     jwks_uri: Option<Url>,
     id_pointer: Option<String>,
-    opa_executor: WasmtimeOpaExecutor,
+    opa: Arc<Mutex<WasmtimeOpaExecutor>>,
```
ryanbtp, line 232: This is fine for now, but we may want to look at moving this to process in spawn_blocking depending on how our profiling turns out

ryanbtp:
For a performance PR, a benchmark is a great thing to add (particularly for something like this that's readily microbenchmarked). Unfortunately bench is still nightly. Have a look at writing a https://github.com/bheisler/criterion.rs test for first / subsequent executions

ryanbtp (approved):

crates/common/src/lib.rs
```diff
@@ -1,4 +1,4 @@
-#![cfg_attr(feature = "strict", deny(warnings))]
+// #![cfg_attr(feature = "strict", deny(warnings))]
```
mtbc, line 1: Do we want to leave this in as-is?

suchapalaver: Agh! No! Thanks for catching that!

mtbc (approved):

PR #184, opened 2023-02-01
guard GraphQL API with OPA

mtbc:
Effect CHRON-211 by having all resolution of GraphQL request elements checked with OPA.


mtbc:
My main difficulties here were working out how to have the endpoints insert the `AuthId` into the GraphQL context (different for each), and figuring out the `opa_executor.lock().await` incantation (using a tokio `Mutex`).

crates/api/src/chronicle_graphql/mod.rs
```diff
@@ -373,37 +374,134 @@ where
...
+            let context = serde_json::to_value(self)?;
+            if opa_executor
+                .lock()
+                .await
+                .evaluate(identity, &context)
+                .await?
+            {
+                Ok(())
+            } else {
+                Err("violation of policy rules".into())
```
ryanbtp, line 415: Can we log something to info with more detail here (identity + context), and #[instrument] the guard method. The error ends up in the client, so keeping that mysterious is good but operationally we need more details here

crates/api/src/chronicle_graphql/mod.rs
```diff
@@ -373,37 +374,134 @@ where
...
                         let span = trace_span!("Execute GraphQL request", request = ?req);
                         return Ok(GraphQLBatchResponse(
                             self.schema.execute_batch(req).instrument(span).await,
-                        ));
+                        )
+                        .into_response());
+                    }
+                }
+            }
+            tracing::warn!("rejected authorization: {:?}", authorization);
```
ryanbtp, line 458: can we use the non string form for events, i.e (rejected_authorisation = ?authorisation), this will be much easier to query for in log aggregators than string formatting. Remind me when I forget this too!

crates/chronicle/src/codegen/mod.rs
```diff
@@ -774,6 +774,7 @@ fn gen_query() -> rust::Tokens {
     impl Query {
 
     #[allow(clippy::too_many_arguments)]
+    #[graphql(guard = "chronicle_graphql::AuthGuard::new(chronicle_graphql::AccessType::QUERY, \"activityTimeline\")")]
```
ryanbtp, line 777: I think we need a more generalised design than this, guards also need to be applied to resolvers as well as methods. Context has everything we need I think:

https://github.com/async-graphql/async-graphql/blob/master/src/context.rs#L250

If we generalise guard to use context internally, then it can also be easily applied to objects, with this approach that's going to be painful

ryanbtp: https://github.com/async-graphql/async-graphql/blob/e839ffafcce3b114289d61a54cb5edbe1377e1a0/tests/guard.rs#L58

See here, for what we need to do - all fields need to be guarded for fine grained query authorisation, not just the root query / mutations

ryanbtp (changes requested):

crates/chronicle/src/codegen/mod.rs
ryanbtp: https://github.com/async-graphql/async-graphql/blob/ad05baf79aea55a0827291abf55af5ce7cc8d386/src/validation/rules/known_directives.rs#L25

```
     self.location_stack
            .push(match &operation_definition.node.ty {
                OperationType::Query => __DirectiveLocation::QUERY,
                OperationType::Mutation => __DirectiveLocation::MUTATION,
                OperationType::Subscription => __DirectiveLocation::SUBSCRIPTION,
            });
            ```

Context has everything we need to make an auth choice already parsed and available, there's no need to pass anything into the guard itself, we just need to get some things from the context and forward it to the opa executor along with the identity

File: crates/api/src/chronicle_graphql/mod.rs (hunk 2 above)
Comment (mtbc) [reply crates/api/src/chronicle_graphql/mod.rs:458]: That `authorization` is just the string from the HTTP header but I'll add a log line for if the credentials got decoded.

Comment (ryanbtp) [reply crates/api/src/chronicle_graphql/mod.rs:458]: Ah no, it's the message I am referring to.

```
"rejected authorization: {:?}", authorization
```

Will come up as "rejected authorisation {literal json}" in a string

rejected_authorisation=?authorisation will emit structure log data, not a string that needs to be parsed

Comment (mtbc) [reply crates/api/src/chronicle_graphql/mod.rs:458]: The `authorization` string just has `Bearer base64stuff....` - but I can try the `?` approach on the `bearer_token` struct if we manage to parse the base64stuff out of it?

Comment (mtbc) [reply crates/api/src/chronicle_graphql/mod.rs:458]: It doesn't look as if the `bearer_token` has JSON in it either. Well, it was worth a look.

File: crates/chronicle/src/codegen/mod.rs (hunk 3 above)
Comment (mtbc) [reply crates/chronicle/src/codegen/mod.rs:777]: Thank you for these tips. My current impression is that we can get rid of those guards and instead hook into `async_graphql::extensions::Extension::resolve`. I shall experiment accordingly.

Comment (mtbc) [reply crates/chronicle/src/codegen/mod.rs]: @ryanbtp: It turns out that using another extension does appear to allow us to avoid touching codegen. The `resolve_path` that I use in the new rego file includes the specific elements that the request tries to query out of the server.

File: crates/api/src/chronicle_graphql/mod.rs [hunk 4]
```diff
@@ -458,6 +523,73 @@ impl async_graphql::extensions::ExtensionFactory for AuthFromJwt {
...
+        info: async_graphql::extensions::ResolveInfo<'_>,
+        next: async_graphql::extensions::NextResolve<'_>,
+    ) -> async_graphql::ServerResult<Option<async_graphql::Value>> {
+        use async_graphql::ServerError;
+        use serde_json::Value;
+        if let (Some(identity), Some(opa_executor)) = (
+            ctx.data_opt::<AuthId>(),
+            ctx.data_opt::<Arc<Mutex<WasmtimeOpaExecutor>>>(),
+        ) {
+            let opa_context = Value::Object(
```
ryanbtp, line 543: We should probably get the operation type in here too, Subscribe / Query / Mutate

ryanbtp (changes requested):

ryanbtp (approved):

PR #185, opened 2023-02-03
Execute OPA rules in Sawtooth TP

suchapalaver:
- [use tagged enum representation on identity object](https://github.com/btpworks/chronicle/commit/c6859fd526154893e6ba54f79f8cde07dc922994)
- [wrap executor in clonable ExecutorContext](https://github.com/btpworks/chronicle/pull/185/commits/c90960c7e418f98950f80855a80cdadc9fae19d0)
- [refactor boostrap of opa executor in graphql server](https://github.com/btpworks/chronicle/pull/185/commits/3770f71eafa483a56f7069ae1f0ea76c239c7fe0)
- [enforce OPA rules in the TP](https://github.com/btpworks/chronicle/pull/185/commits/15c5295f71e356b5e3de0c6f1b52ccefdf1465c2)

NOTES:
- Assume the OPA rule itself is only loaded from the filesystem / embed for now - bootstrapping `ChronicleTransactionHandler` from an embedded `Chronicle` identity is authorized policy
- Failure to initialize `ChronicleTransactionHandler` panics in `sawtooth-tp/main`

[CHRON-220](https://blockchaintp.atlassian.net/browse/CHRON-220)

[CHRON-220]: https://blockchaintp.atlassian.net/browse/CHRON-220?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


crates/sawtooth-tp/src/tp.rs
```diff
@@ -21,20 +26,46 @@ use tracing::{debug, error, info, instrument};
...
+
+fn cli_policy_loader_from_embedded(wasm: &str, entrypoint: &str) -> CliPolicyLoader {
+    if let Some(file) = DefaultAllowOpaPolicy::get(wasm) {
+        let policy = file.data.to_vec();
+        let mut loader = CliPolicyLoader::new();
+        loader.set_entrypoint(entrypoint);
+        loader.load_policy_from_bytes(&policy);
+        loader
+    } else {
+        panic!("TP failed to initialize Cli policy loader")
```
suchapalaver, line 41:
```suggestion
        panic!("Better error message")
```

crates/sawtooth-tp/src/tp.rs
```diff
@@ -216,6 +250,71 @@ impl TP for ChronicleTransactionHandler {
...
+            .dependencies()
+            .into_iter()
+            .collect::<HashSet<_>>()
+            .iter()
+            .map(SawtoothAddress::from)
+            .collect::<HashSet<_>>();
+
+        let operation_state = state.opa_context(deps);
+
+        let state = tx
```
suchapalaver, line 291: Should this perhaps be `spawn_blocking`?

crates/common/src/opa_executor.rs
```diff
@@ -297,9 +328,29 @@ impl OpaExecutor for WasmtimeOpaExecutor {
...
+    }
+}
+
+pub trait EvaluationContext {
+    /// Transform the @graph of an operations dependencies in a ProvModel
+    /// into a more useful representation for looking up from an operation
+    fn format_context_data(&mut self) -> Result<serde_json::Value, ProcessorError>;
+}
+
+impl EvaluationContext for serde_json::Value {
```
ryanbtp, line 345: Hmm, I am not sure I like this being on Value, can we use a newtype wrapper or some sort that indicates it's a serialized / compacted ProvModel

ryanbtp (changes requested):

crates/common/src/ledger.rs
```diff
@@ -880,7 +880,17 @@ impl ChronicleOperation {
                 if let Some(graph) = json_ld.get_mut("@graph").and_then(|g| g.as_array_mut()) {
                     graph
                         .iter_mut()
-                        .map(|entry| entry.format_context_data())
+                        .map(|entry| {
```
suchapalaver, line 883: Duplicating the logic of `EvaluationContext::format_context_data` here to avoid owning each `entry`. Meanwhile, `EvaluationContext::format_context_data` allows us to clearly test how we're transforming the compacted prov model graph data with a view to experimentation down the line.

crates/common/src/opa_executor.rs
suchapalaver: I scrapped the trait. At first I thought I could use a newtype but implementing derefmut, according to the docs, should only be for smart pointers.

crates/sawtooth-tp/src/tp.rs
```diff
@@ -216,6 +255,62 @@ impl TP for ChronicleTransactionHandler {
...
+        let context = serde_json::json!({
+            "identity": identity_context,
+            "operation": operation,
+            "state": state,
+        });
+
+        info!(opa_evaluation_context = %context);
+
+        match opa_executor.evaluate(&identity, &context).await {
+            Ok(_) => Ok(()),
```
mtbc, line 310: I am most curious that an `Ok(false)` is just fine here, we're just checking that we _can_ evaluate it.

suchapalaver: Thanks for asking about that. The updated `evaluate` method returns an error if the result of the evaluation is false, so an `Ok` `Result` here will always be `Ok(true)`. Any suggestions on how to make that less confusing?

suchapalaver: I've added a comment for clarification.

mtbc: I'd suggest returning `Ok(())` - the `bool`'s become superfluous so that's why it's confusing.

crates/api/src/chronicle_graphql/mod.rs
```diff
@@ -560,23 +559,13 @@ impl async_graphql::extensions::Extension for OpaCheck {
...
-                        identity,
-                        opa_context
-                    );
-                    Err(ServerError::new("violation of policy rules", None))
-                }
+                Ok(()) => next.run(ctx, info).await,
                 Err(error) => Err(ServerError {
-                    message: "cannot check policy rules".to_string(),
+                    message: format!(
+                        "attempt to violate policy rules by {identity} in context {opa_context:#?}"
```
mtbc, line 567: It seemed safer to me to keep the internal state out of the `ServerError::message` that gets sent to the remote user, instead just keeping it to the logs so only the admins see the server's details of attempted violations.

mtbc (changes requested):

crates/api/src/chronicle_graphql/mod.rs
suchapalaver: Thanks for catching this!

mtbc (approved):

ryanbtp (dismissed):

ryanbtp (approved):

PR #190, opened 2023-02-10
JWT identity variant containing struct with scalar and JWT claims

suchapalaver:
[CHRON-234](https://blockchaintp.atlassian.net/browse/CHRON-234)


crates/common/src/identity.rs
```diff
@@ -35,8 +55,17 @@ impl TryFrom<&str> for AuthId {
...
-        Self::JWT(agent.to_owned())
+    pub fn agent(claims: &JwtClaims, json_pointer: &str) -> Result<Self, IdentityError> {
+        let claims = claims.to_owned();
+
+        if let Some(serde_json::Value::String(external_id)) =
+            serde_json::Value::Object(claims.0.clone()).pointer(json_pointer)
+        {
+            Ok(Self::JWT(JwtId::new(external_id, claims)))
+        } else {
+            warn!("Invalid JWT claims: {:#?}", claims.0);
```
mtbc, line 66: It may be instead that the JSON pointer is wrong, maybe instead an error about not being able to get the agent ID from the claims?

mtbc:
This looks good to me though, from the ticket, I'm most curious which group policy code doesn't already have access to the claims anyway, maybe something in the depths of the transaction processing needs something other than OPA?

ryanbtp:
Can we also rework the claims / OPA context creation in graphql to use the same types and ensure the identity is used in the same way by rego? Otherwise we end up with 2 ways of writing OPA rules

crates/api/src/chronicle_graphql/mod.rs
```diff
@@ -373,9 +379,6 @@ where
     }
 }
 
-#[derive(Clone, Debug)]
-pub struct JwtClaims(pub serde_json::Map<String, serde_json::Value>);
-
```
suchapalaver, line 378: Move this to identity.rs to resolve cyclic dependency otherwise

crates/api/src/chronicle_graphql/mod.rs
```diff
@@ -551,15 +558,9 @@ pub struct AuthFromJwt {
...
-        use common::prov::AgentId;
-        use serde_json::Value;
-
-        if let Some(Value::String(external_id)) =
-            Value::Object(claims.0.clone()).pointer(&self.json_pointer)
-        {
-            Some(AuthId::agent(&AgentId::from_external_id(external_id)))
-        } else {
-            None
+        match AuthId::from_jwt_claims(claims, &self.json_pointer) {
```
mtbc, line 561: is this simply `AuthId::from_jwt_claims(claims, &self.json_pointer).ok()` or somesuch?

suchapalaver: What do you mean, sorry?

suchapalaver: Actually, I like it, I've changed it. Thanks

crates/common/src/identity.rs
```diff
@@ -1,29 +1,116 @@
...
+
+/// Chronicle identity object for authorization
 #[derive(Serialize, Deserialize, PartialEq, Eq, Debug, Clone)]
-#[serde(tag = "type", content = "id")]
+#[serde(tag = "type")]
 pub enum AuthId {
-    Anonymous,
-    Chronicle,
-    JWT(AgentId),
+    Anonymous(serde_json::Value),
```
mtbc, line 101: Curious what these Values are for. Do we ever use them?

mtbc:
No issues from my point of view, happy to see this go forward. :+1:

crates/common/src/identity.rs
suchapalaver: Is your sense that context data for OPA checks is irrelevant in all cases to Chronicle and Anonymous users?

mtbc: I just couldn't find where we put any actual context data in for those users but I may have just missed the code.

ryanbtp: Context is the WHAT of an operation, Identity is the WHO. 

The WHO will be the same in both graphql and the transaction processor but the WHAT is different - in graphql it's whatever async_graphql thinks we are trying to do via context, in the transaction processor it is the chronicle operation (which  has already passed and doesn't need to be part of the WHO).

ryanbtp: So here anonymous is just anonymous, the context is separate as it is dependent on where it is being checked.

crates/common/src/identity.rs
```diff
@@ -1,29 +1,116 @@
...
+        }
+    }
+}
+
+/// Contains the scalar ID and OPA contextual data for a user established via JWT
+#[derive(Serialize, Deserialize, PartialEq, Eq, Debug, Clone)]
+pub struct JwtId {
+    pub id: AgentId,
+    #[serde(flatten)]
+    pub context: serde_json::Value,
```
ryanbtp, line 85: This would be better called claims, I think this is how we got confused

ryanbtp: Neither anonymous or chronicle need an inner type, they should just be standalone

ryanbtp (changes requested):

.github/workflows/cargo-checkmate.yaml
```diff
@@ -10,6 +10,10 @@ jobs:
       - name: "Install protoc"
         run: |
           sudo apt-get install -y protobuf-compiler
+      - name: "Setup OPA"
```
mtbc, line 13:
```suggestion
      - name: "Set up OPA"
```

mtbc, line 13:
```suggestion
      - name: "Set up OPA"
```

crates/chronicle/src/bootstrap/mod.rs
```diff
@@ -333,9 +313,11 @@ where
             }
         }
 
-        if jwks_uri.is_none() && id_pointer.is_none() {
+        // This may not be the ideal place for this, but we need it for development
+        // as we default to an anonymous user when we serve graphql
+        if jwks_uri.is_none() {
             DirectoryStoredKeys::new(config.secrets.path)?
-                .generate_agent(&AgentId::from_external_id("Anonymous"))?;
+                .generate_agent(&AgentId::from_external_id("anonymous"))?;
```
mtbc, line 320: I've been curious if identities generated like this are materially distinct from `AuthId::anonymous()`.

suchapalaver: This generates a user key.

crates/chronicle/src/bootstrap/mod.rs
```diff
@@ -313,13 +313,10 @@ where
...
         };
 
-        let id_pointer = if matches.is_present("anonymous-api") {
-            None
-        } else {
-            matches
-                .value_of("id-pointer")
-                .map(|id_pointer| id_pointer.to_string())
-        };
+        // Defaults to look for "/sub"
```
mtbc, line 316:
```suggestion
        // CliModel defaults to look for "/sub"
```
Something like that might help piece things together if the code in `cli.md` is changed by somebody unaware of this comment.

ryanbtp (approved):

PR #191, opened 2023-02-13
allow specification of required JWT claims

mtbc:
In researching JWT contents for CHRON-230, I came across consensus that specific claims are well worth checking for expected values, so this PR includes ability to have Chronicle check them accordingly before accepting a JWT. This work is enough separable from the main work for the ticket.

So, for instance, in testing with JWTs from Auth0, one might set `JWT_MUST_CLAIM_ISS` to one's Auth0 tenancy and `JWT_MUST_CLAIM_AUD` to the application API created therein. I also set `JWT_MUST_CLAIM_AZP` to the application's client ID. This is part of ensuring that the JWT was created for the use of Chronicle.

I avoided making any specific check mandatory, to cover for variations in deployment environment.

### DIFF
--- a/crates/api/src/chronicle_graphql/mod.rs
+++ b/crates/api/src/chronicle_graphql/mod.rs
@@ -30,7 +30,7 @@ use poem::{
     Endpoint, Route, Server,
 };
 use serde::{Deserialize, Serialize};
-use std::{net::SocketAddr, sync::Arc};
+use std::{collections::HashMap, net::SocketAddr, sync::Arc};
 use tokio::sync::broadcast::error::RecvError;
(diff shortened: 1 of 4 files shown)


mtbc:
Part of my thinking in separating these from OPA policy rules, aside from performance, was that they are about if we believe the claims at all, whatever they are, and they may vary by authentication provider configuration, so environment variables will make it easier for the integrators to adjust for that. I was more comfortable with propagating the claims through various layers into Chronicle only if we already believed them, regardless of what they authorize the user to do.

ryanbtp (approved):

PR #193, opened 2023-02-15
allow JWT claims to be fleshed out using /userinfo endpoint

mtbc:
Enables optional `--userinfo-address` option which switches behavior to presenting provided bearer token to that endpoint and taking claims as union of those from original token (if not opaque) and those from endpoint (either JWT or plain JSON). Enables CHRON-230 in that scopes may be in access token but user information for deriving Chronicle ID may need requesting separately.

If merged, I'll open a followup ticket for caching the user info so we don't hit the endpoint for every request.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -125,6 +125,7 @@ dependencies = [
  "r2d2",
  "rand 0.8.5",
  "rand_core 0.6.4",
+ "reqwest",
  "serde",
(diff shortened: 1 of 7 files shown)


mtbc:
For testing this, one may use the *access* token from the code commented on CHRON-236.

ryanbtp (approved):

PR #194, opened 2023-02-15
Correctly pass ID from JWT claims to OPA evaluator

mtbc:
Fixes CHRON-236.

Data set on the GraphQL request in `prepare_request` reaches the handler code's context just fine but not `resolve`'s extension context. Fortunately, for determining a Chronicle ID, it is easy enough to have `OpaCheck` try the JSON pointer separately.

### DIFF
--- a/crates/api/src/chronicle_graphql/mod.rs
+++ b/crates/api/src/chronicle_graphql/mod.rs
@@ -542,8 +542,25 @@ where
     }
 }
 
-struct AuthFromJwt {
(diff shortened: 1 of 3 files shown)


mtbc:
The included instrumenting of the OPA `evaluate` is helpful for testing.

mtbc:
Use the ID token from the code commented on CHRON-236 to generate a JWT for testing here.

ryanbtp (approved):

PR #195, opened 2023-02-16
cache responses from OIDC /userinfo endpoint

mtbc:
Resolves CHRON-237 by adding caching of responses from when with `--userinfo-address` we pass an access token to an authorization server endpoint to obtain user information for determining the Chronicle ID.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -101,6 +101,7 @@ dependencies = [
  "async-stream",
  "async-trait",
  "base64 0.21.0",
+ "cached",
  "chronicle-telemetry",
  "chrono",
  "common",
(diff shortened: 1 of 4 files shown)


mtbc:
This PR should cover the common cases well enough, for example, where we don't use `/userinfo` or where it helps us to flesh out the claims.

There are a few other cases to consider. For example, with caching, a couple of situations:
* This *does not* cache where `/userinfo`s response with further data was unintelligible or failed verification.
* This *does* cache where we verified claims from the initial token but `/userinfo`s response was not HTTP 200.

We might should pursue a followup ticket / PR for examining the edge cases and deciding caching for each.

A complication is that, if providing error sources, it is difficult to provide cached ones because they may not be `Clone`able.

PR #196, opened 2023-02-17
fix(docker): correct sawtooth genesis

scealiontach:
Fixes CHRON-228

Signed-off-by: Kevin O'Donnell <kevin.odonnell@btp.works>

PR #199, opened 2023-02-17
fix: remove opa cli args and use default allow opa rule

suchapalaver:


scealiontach:
Force merging this so that #196  can be merged, fixing the build

PR #200, opened 2023-02-18
Feature/build-and-test-opa

ryanbtp:
* Test container, using opactl to control the opa TP
* Currently integrated into chronicle-test, but we will split this out
  when moving crates

Signed-off-by: Ryan Roberts <ryan.roberts@btp.works>

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -36,6 +36,7 @@ pipeline {
       steps {
         sh '''
           make clean build
+          make -f opa.mk clean build
         '''
       }
     }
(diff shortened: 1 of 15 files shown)


docker/unified-builder
```diff
@@ -150,3 +154,15 @@ FROM debian:bullseye-slim AS chronicle-tp
 ARG TARGETARCH
 WORKDIR /
 COPY .artifacts/artifacts/${TARGETARCH}/chronicle_sawtooth_tp /usr/local/bin
+
```
scealiontach, line 157: Since opa-tp is destined to become a separate repo, it is best if this were a full separate builder

scealiontach (changes requested):

docker/unified-builder
ryanbtp: Ok, I was going to split that out in when needed as it still ends up in the makefile, but can do this now

suchapalaver (dismissed):

scealiontach (dismissed):

suchapalaver (approved):

suchapalaver (approved):

suchapalaver (dismissed):

suchapalaver (approved):

PR #203, opened 2023-02-21
retry failed database connection

mtbc:
Retries the PostgreSQL connection, with increasing backoff, then gives up and continues as it would have previously if facing an immediate failure. Resolves CHRON-233.

Developers may prefer to give the `--embedded-database` option to skip the retries and waits.


suchapalaver:
Maybe add that useful note to our docs?
> Developers may prefer to give the --embedded-database option to skip the retries and waits.

mtbc:
> Maybe add that useful note to our docs?

Good thought! Fortunately, I think it's probably enough there in the last part of `config.md`. We should expand `cli.md` in general at some point though.

suchapalaver (approved):

PR #205, opened 2023-02-23
retry after PgEmbedError PgStartFailure

mtbc:
To resolve CHRON-248 this generalizes the `pool_remote` retry code into a trait that is also then used before constructing the embedded database pool.


suchapalaver:
This looks good to me, but would be even better for @ryanbtp to approve this.

mtbc:
A couple of CI runs seem to confirm that the first commit necessitates the second, I don't know why.

ryanbtp (approved):

PR #209, opened 2023-02-27
fix: unlock cors with --open

suchapalaver:

[CHRON-252](https://blockchaintp.atlassian.net/browse/CHRON-252)


crates/api/src/chronicle_graphql/mod.rs
```diff
@@ -741,6 +764,13 @@ where
             }
         };
 
+        if sec.open {
+            tokio::spawn(async move {
+                tokio::time::sleep(Duration::from_millis(200)).await;
```
mtbc, line 769: Is this delay always going to be good and enough? I guess there's no easy way we can simply wait for whatever this delay is for?

crates/api/src/chronicle_graphql/mod.rs
```diff
@@ -741,6 +764,13 @@ where
...
+                open::that(format!("http://{}", address)).ok();
```
mtbc, line 770: Are we okay to throw away any `Err` here?

crates/chronicle/src/bootstrap/cli.rs
```diff
@@ -1003,6 +1003,12 @@ impl SubCommand for CliModel {
                             .takes_value(true)
                             .default_value("127.0.0.1:9982")
                             .help("The graphql server address (default 127.0.0.1:9982)"),
+                    ).arg(
+                        Arg::new("open")
```
mtbc, line 1007: Does it make enough sense that an "unlock CORS" option is called `--open`?

mtbc:
Looks good overall! I have a couple of small questions but don't let them block progress here, none of them are showstoppers, just more possible food for thought. If @ryanbtp approves just go ahead, resolve and merge.

crates/api/src/chronicle_graphql/mod.rs
suchapalaver: For now, simply replacing what was removed.

crates/api/src/chronicle_graphql/mod.rs
suchapalaver: As I say, this is just replacing what we once had so that things work, for now.

crates/chronicle/src/bootstrap/cli.rs
suchapalaver: Great point, but for now it would mean making changes to other repositories.

mtbc: "for now" - probably also subsequently, or is there a coming change that'll make it easier? My concern is that it will become further entrenched by more people/scripts using it, thus increasingly much pain to change, so now may be our best opportunity, or we decide it won't cause enough confusion to be worth it. If we're not sure, or for some reason it'll be easier later, fine with me if we just ask in scrum, open a ticket raising the question, etc. rather than impeding this PR.

It's also worth keeping an eye out for possibilities to improve development workflows to make multi-repository changes less inhibiting.

suchapalaver: All great points, this is just more about unbreaking things. And even better if it raises questions about what we want from this after we get things working again.

ryanbtp: We could alias the command to make migration easier

ryanbtp: That way we can take out --open once deps are updated, similarly to how we handle deprecation in library apis

mtbc: I'll move them to a ticket. Just trying to prevent us from unbreaking things in ways we come to regret!

suchapalaver: Thanks @ryanbtp , I hadn't looked at aliases in clap before. `--unlock-cors` sound good to you as an alias, @mtbc ?

mtbc: Sure, my point is more that we should think about this stuff sooner rather than later: ideally we can name/alias it now something that we're happy with long-term but it's fine to get this PR in first. Feel free to defer all this to when we pick up CHRON-254.

As it *does* open something, it might also help simply to explain a bit more in the longer help text. Personally, I would have been surprised to find what's simply an "Unlock CORS" option opening a "method not allowed" browser tab at me but it's been a while since I had to focus on web security.

I guess it depends what it is *doing*. How does opening a browser tab unlock CORS? Perhaps there's a succinct statement there that can inspire the name or description. I'm fine for that to go into a later PR (or for us to decide that `--open` is fine) just so long as we get around to it before everyone's gotten used to the current.

suchapalaver: It turned out that not least of the confusion here was being caused by having not put back all the parts necessary to do what the PR says, but that's all there now and we're back with the original functionality of being able to use the graphql playground.

suchapalaver: @mtbc I've also documented what this is useful for now.

ryanbtp (approved):

PR #216, opened 2023-03-03
do not suggest support for wasAttributedTo

mtbc:
Both @ChrisH1066 and I came away from the starting points diagram and "Chronicle extends the core PROV-O vocabulary" assuming that `wasAttributedTo` is offered, we probably won't be the last. This addition succinctly forestalls that misapprehension, can be reverted once CHRON-124 is done.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ for testing purposes.
 ![PROV-O](https://www.w3.org/TR/prov-o/diagrams/starting-points.svg)
 
 Chronicle extends the core PROV-O vocabulary as described
-[here](docs/chronicle_vocabulary.md).
+[here](docs/chronicle_vocabulary.md), except for `wasAttributedTo`.
 
 ## Deployment
 


README.md
```diff
@@ -29,7 +29,7 @@ for testing purposes.
 ![PROV-O](https://www.w3.org/TR/prov-o/diagrams/starting-points.svg)
 
 Chronicle extends the core PROV-O vocabulary as described
-[here](docs/chronicle_vocabulary.md).
+[here](docs/chronicle_vocabulary.md), except for `wasAttributedTo`.
```
scealiontach, line 32: You can just define a wasAttributedTo property in the domain.yaml, no?  And besides even if you can't d that, I'd rather we spent the effort and actually put the implementation in place in the core.

scealiontach (changes requested):

README.md
mtbc: We don't support Agent property values but I suppose one could put the agent ID as a string then run a followup query. At least, after looking at this README, we were rather expecting a corresponding mutation to be offered.

I'm all for putting the implementation in place in the core, can close this and add a to-do ticket for that instead? This doc warning was just intended to linger until that had happened as it wasn't obviously near-term.

ryanbtp: It's a pretty simple implementation, we should also transitively return agent -> entity via activity right? Its use seems pretty specialised to when you categorically do not know the activity involved

mtbc: Maybe also for when the indirection via activity doesn't add anything non-obvious or interesting for the problem, or want to add something more direct for easier querying.

The idea of defining a property makes me wonder, what if we could define properties that take entities, agents, activities, as values? But maybe that's very much putting the cart before the horse. Next we'll be wanting properties that are dictionaries of agents to lists of booleans or somesuch.

This PR is just about not having more of us expect a `wasAttributedTo` that turns out not to be offered. Looks like the possibilities for `wasAttributedTo` are,

1. already enough there that as-is suffices
1. maybe not there anytime soon so good not to suggest otherwise
1. not there now but will be soon.

PR #223, opened 2023-03-10
Feature/chronicle uses client sdk

ryanbtp:


crates/async-sawtooth-sdk/src/error.rs
```diff
@@ -0,0 +1,45 @@
+use futures::channel::mpsc::SendError;
+use protobuf::ProtobufError;
+use sawtooth_sdk::messaging::stream::ReceiveError;
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum SawtoothCommunicationError {
+    #[error("ZMQ error {0}")]
```
mtbc, line 8:
```suggestion
    #[error("ZMQ error: {0}")]
```
If touching this anyway, some colons help with readability.

crates/chronicle-protocol/Cargo.toml
```diff
@@ -0,0 +1,63 @@
...
+  "pem",
+] }
+lazy_static = "1.4.0"
+openssl = "0.10.38"
+prost = "0.10.0"
+prost-types = "0.10.0"
+rand = { version = "0.8.5", features = ["getrandom"] }
+rand_core = "0.6.3"
+sawtooth-sdk = "0.5.2"
+serde = "1.0.136"
```
mtbc, line 40:
```suggestion
serde.workspace = true
```
and much similar?

crates/common/src/prov/model/from_json_ld.rs
```diff
@@ -54,7 +54,7 @@ impl Loader<IriBuf, ()> for ContextLoader {
             let json = json!({
              "@context": crate::context::PROV.clone()
             });
-            let value = json_ld::syntax::Value::from_serde_json(json, |_| ());
+            let value = json_syntax::Value::from_serde_json(json, |_| ());
```
mtbc, line 57: Is this why we now depend on `json-syntax`? Odd it's for just this `from_serde_json` usage, there's a couple more.

crates/api/src/inmem.rs
```diff
@@ -0,0 +1,357 @@
...
+
+    pub fn readable_state(&self) -> Vec<(String, Value)> {
+        // Deal with the fact that policies are raw bytes, but meta data and
+        // keys are json
+
+        self.state
+            .borrow()
+            .iter()
+            .map(|(k, v)| {
+                let as_string = String::from_utf8(v.clone()).unwrap();
```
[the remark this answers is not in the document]
ryanbtp: This is test double code - so it is good if it does blow up in that case

crates/async-sawtooth-sdk/src/error.rs
ryanbtp: Good spot

mtbc: Couple of other similar touched in other files, we'll get them all in the end though!

mtbc:
Fixed for a Mac issue but now investigating a build issue, should be an easy fix.

crates/async-sawtooth-sdk/src/error.rs
mtbc: Maybe the others in the same block at least?

Cargo.toml
```diff
@@ -4,46 +4,47 @@ members = ["crates/*"]
...
+futures = "0.3.21"
+genco = "0.16.1"
+glob = "0.3.0"
+hashbrown = "0.13"
+hex = "0.4.3"
+insta = { version = "1.26.0", features = ["redactions", "toml"] }
+iref = "2.2"
+iref-enum = "2.1"
+json-ld = { version = "0.14" }
+json-syntax = { version = "*", features = ["serde", "serde_json"] }
```
mtbc, line 44: Can this now be removed?

ryanbtp: No, this is how the feature is enabled in a sub dependency

mtbc: How come we need it now but didn't before?

ryanbtp: Unknown, but this is the only way to activate features in transitive dependencies I know of

mtbc: Thanks. Most curious that we didn't need this before this PR, do let me know if you ever figure out what the relevant trigger was!

crates/api/src/lib.rs
```diff
@@ -1539,9 +1508,10 @@ mod test {
     struct TestDispatch<'a> {
         api: ApiDispatch,
         _db: TemporaryDatabase<'a>, // share lifetime
+        _tp: EmbeddedChronicleTp,
     }
 
-    impl TestDispatch<'_> {
+    impl<'a> TestDispatch<'a> {
```
mtbc, line 1514: Feels like a no-op but no objection.

ryanbtp: It is, I'll clippy it out after this is in main

PR #224, opened 2023-03-10
Add documentation capabilities to chronicle domain definition

suchapalaver:
- [x] add documentation capabilities for non-domain-specific items that become part of GraphQL schema
- [x] add docs capabilities for domain-specific items

If you want to see how the generated docs look in a GraphQL client and you have @btpworks credentials, clone `btpworks/chronicle-examples`, checkout [feature/in-domain-documentation-for-artworld](https://github.com/btpworks/chronicle-examples/tree/feature/in-domain-documentation-for-artworld), and run the `Artworld` domain.

---
[CHRON-145](https://blockchaintp.atlassian.net/browse/CHRON-145)

[CHRON-145]: https://blockchaintp.atlassian.net/browse/CHRON-145?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


crates/api/src/chronicle_graphql/mod.rs
```diff
@@ -134,6 +143,11 @@ pub struct Namespace {
 }
 
 #[Object]
+/// # chronicle:Namespace
```
mtbc, line 146:
```suggestion
/// # `chronicle:Namespace`
```

crates/api/src/chronicle_graphql/mod.rs
```diff
@@ -72,6 +72,11 @@ pub struct Identity {
 }
 
 #[Object]
```
suchapalaver, line 74: I haven't figured out how to use `include_str`the way I've been able to in `codegen` for these types that get included in GraphQL schema.

crates/chronicle/src/bootstrap/mod.rs
```diff
@@ -640,31 +640,31 @@ pub mod test {
     fn test_cli_model() -> CliModel {
         CliModel::from(
             ChronicleDomainDef::build("test")
```
suchapalaver, line 642: `None` here is the `doc` field (`Option<String>`).

crates/chronicle/src/codegen/mod.rs
```diff
@@ -63,12 +70,26 @@ fn gen_type_enums(domain: &ChronicleDomainDef) -> rust::Tokens {
     let graphql_enum = &rust::import("chronicle::async_graphql", "Enum");
     let domain_type_id = &rust::import("chronicle::common::prov", "DomaintypeId");
     let prov_role = &rust::import("chronicle::common::prov", "Role").qualified();
-    quote! {
```
suchapalaver, line 73: It would perhaps be tidier to have `domain_docs/` inside `docs/`, but currently we `.dockerignore` the latter.

crates/chronicle/src/codegen/model.rs
```diff
@@ -1221,52 +1366,69 @@ pub mod test {
         name: evidence
         attributes:
           CMSId:
+            doc: ~
```
mtbc, line 1369: Bit tedious but if we're feeding in all these doc strings anyway, maybe make them different (but still perfunctory) to make sure that none get mixed up somehow in the output - generally a good idea with tests to vary input values where that can cheaply increase specificity.

crates/common/src/prov/id/mod.rs
```diff
@@ -759,6 +759,8 @@ impl From<&EntityId> for IriRefBuf {
     }
 }
```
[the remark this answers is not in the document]
mtbc: No objections here, we can rethink if it becomes clear that your issue is regarded as won't-fix.

crates/chronicle/src/codegen/model.rs
suchapalaver: These are `None` though, because they're representing a domain model without any doc strings added.

crates/common/src/prov/id/mod.rs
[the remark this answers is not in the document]
ryanbtp: Yes I think something is better than nothing here, and may eventually be resolved properly upstream

crates/chronicle/src/codegen/model.rs
mtbc: Aha, thank you, I didn't realize that `~` signifies `None`.

crates/api/src/chronicle_graphql/mod.rs
ryanbtp: Does desc / description work in the macro body rather than as a comment? https://docs.rs/async-graphql/latest/async_graphql/attr.Object.html

crates/chronicle/src/codegen/mod.rs
ryanbtp: I think it's reasonable to keep them separate here

crates/api/src/chronicle_graphql/mod.rs
suchapalaver: I also tried the following but I get the same `expected literal` error that I haven't been able to figure out.

```rust
#[Object]
#[graphql(desc = include_str!("../../../../domain_docs/identity.md"))]
impl Identity {
    async fn public_key(&self) -> &str {
        &self.public_key
    }
}
```

The following results in the same issue, whereas it works on "normal" Rust code: 

```rust
#[Object]
#[doc = include_str!("../../../domain_docs/identity.md")]
impl Identity {
    async fn public_key(&self) -> &str {
        &self.public_key
    }
}
```

Experimenting with the place of the `///` doc comment showed it would only show up in the GraphQL schema where it is now.

suchapalaver: Same if you do that:

```rust
#[Object(desc = include_str!("../../../../domain_docs/association.md"))]
```

ryanbtp: Annoying!

ryanbtp (approved):

PR #227, opened 2023-03-15
fix(chronicle_graphql): do not open browser when unlocking playground

suchapalaver:
Remove arbitrary wait and opening the browser with `--unlock-cors`/`--open` option.
Change existing CLI flag to be less confusing, changing it from `--unlock-cors` to `--playground`

This was sometimes failing with `chronicle-examples` and therefore is something we want to do before next release.

---


scealiontach (changes requested):
Is it possible to improve this to an `--allow-origin foo,bar,` or `--allow-origin *`  parameter rather than a boolean?

PR #228, opened 2023-03-15
Do not log user info

suchapalaver:
- Customize `Debug` implementation for 
    - `Config`
    - `DirectoryStoredKeys`
    - `KeyImport`
    - `JwtClaims` and `JwtId`
- Skip client info instrumentation but include errors in `authorization`  

---
[CHRON-265](https://blockchaintp.atlassian.net/browse/CHRON-265)


mtbc:
I'd like to suggest that we should log the JWT claims as they oughtn't contain any secret authentication material and they may be most useful in debugging OPA policies.

crates/api/src/chronicle_graphql/mod.rs
```diff
@@ -806,7 +862,7 @@ where
             let app = match sec.jwks_uri {
                 Some(jwks_uri) => {
                     const CACHE_EXPIRY_SECONDS: u32 = 100;
-                    tracing::debug!("API endpoint authentication uses {}", jwks_uri);
+                    tracing::debug!("API endpoint authentication uses {:?}", jwks_uri);
```
mtbc, line 865:
```suggestion
                    tracing::debug!("API endpoint authentication uses {jwks_uri:?}");
```
while we're here anyway.

scealiontach:
> I'd like to suggest that we should log the JWT claims (as perhaps augmented via `/userinfo`) as they oughtn't contain any secret authentication material and they may be most useful in debugging OPA policies. Certainly happy to defer to @scealiontach on this question though should more caution be warranted.

Just no keys and secrets of any shape or form.  And nothing that can be construed as PII above trace level, and even then only where *most* needed not everywhere, you can figure out anything else via the span id.  Otherwise fair-game.  But still try to keep the noise low.  spans are there for a reasn.

suchapalaver:
> I'd like to suggest that we should log the JWT claims (as perhaps augmented via `/userinfo`) as they oughtn't contain any secret authentication material and they may be most useful in debugging OPA policies. Certainly happy to defer to @scealiontach on this question though should more caution be warranted.
> 
> _ETA:_ I also wonder if maybe we do want to log `DirectoryStoredKeys`? Again may help in debugging deployment/integration and probably isn't terribly secret.

Dropped the `DirectoryStoredKeys` commit.

crates/api/src/chronicle_graphql/mod.rs
```diff
@@ -413,9 +413,65 @@ where
...
+    pub fn new(uri: Url) -> Self {
+        Self { uri }
+    }
+}
+
+impl std::fmt::Debug for JwksUri {
+    fn fmt(&self, fmt: &mut std::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
+        write!(
+            fmt,
+            r#"JwksUri {{ uri: Url {{ scheme: {}, cannot_be_a_base: {}, username: {}, password: ***SECRET***, host: {:?}, port: {:?}, path: {}, query: {:?}, fragment: {:?} }} }}"#,
```
mtbc, line 430: I wonder if these need more `{:?}`, seems weird to see the only somewhat the usual debug format but then many bare strings, especially when they're empty.

suchapalaver: It's by design as those items do not implement `Debug`, but implement `Display`.

mtbc:
The (encoded) JWT claims are coming through via things like the authorization endpoints' `req` parameter to `respond` or `call`, generally have to watch for where we might report HTTP headers.

Also around that part of the code, better bump the logging of `req.remote_addr()` up to TRACE as IP addresses can be identifying under GDPR.

crates/api/src/chronicle_graphql/mod.rs
```diff
@@ -569,7 +569,7 @@ where
     M: ObjectType + 'static,
     S: SubscriptionType + 'static,
 {
-    #[instrument(level = "debug", skip(self, prepare_req), ret(Debug))]
+    #[instrument(level = "trace", skip(self, prepare_req), ret(Debug))]
```
mtbc, line 572: Might instead want to just skip the `req` for this and below, as `req` effectively includes the JWT claims.

crates/chronicle/src/bootstrap/config.rs
```diff
@@ -127,3 +153,33 @@ address = "{}"
...
+
+    #[test]
+    fn secret_config_custom_debug() {
+        let secret_config = SecretConfig {
+            path: "some_path".into(),
+        };
+
+        insta::assert_debug_snapshot!(secret_config, @r###"
+        SecretConfig {
+            path: "***SECRET***",
```
mtbc, line 172: Sorry I didn't catch this before but would it be better to log this? The path itself isn't secret and can help point to config issues.

ryanbtp: Agreed, very little more annoying than 'file not found' without mentioning the file if not absolutely necessary

crates/api/src/chronicle_graphql/mod.rs
```diff
@@ -582,7 +638,7 @@ where
     M: ObjectType + 'static,
     S: SubscriptionType + 'static,
 {
-    #[instrument(level = "debug", skip(self), ret(Debug))]
+    #[instrument(level = "trace", skip(self), ret(Debug))]
```
mtbc, line 641: We should still be skipping the `req` too as it can contain the JWT claims.

mtbc (approved):

PR #236, opened 2023-03-20
remove dependency on unmaintained json crate

mtbc:
Resolves CHRON-221 by upgrading to current `json-ld` crate to allow removal of `json` crate.

A few bugs in this were found only through automated tests revealing the need for further code updates that static compiler checks could not detect. So, where it seems a code update might warrant a similar or corresponding change elsewhere, check that I made that too.


crates/common/src/prov/model/to_json_ld.rs
```diff
@@ -1047,40 +1050,48 @@ trait Operate {
     fn derivation(&mut self, typ: &DerivationType);
 }
 
-impl Operate for JsonValue {
+impl Operate for Value {
```
suchapalaver, line 1053: I'm surprised you kept this, not sure I would have implemented my own trait on a library type if I came at it now.

mtbc: Good point, I don't know what's idiomatic. I thought it an interesting approach at least, it took me a moment to realize what was happening.

I figured best to focus this PR on the changes that the library upgrade necessitated, we can refactor separately, this was already getting overly large.

crates/common/src/prov/model/mod.rs
```diff
@@ -1163,29 +1164,115 @@ impl ProvModel {
...
-    JsonLd{inner: String}                  = "JsonLd {inner}", //TODO: contribute Send to the upstream JsonLD error type
-    Join{source : JoinError}               = "Tokio",
-    Serde{source: serde_json::Error }      = "Serde conversion {source}",
+    JsonLd{inner: String}                  = "JSON-LD: {inner}", //TODO: contribute Send to the upstream JsonLD error type
+    Join{source : JoinError}               = "Tokio: {source}",
+    Serde{source: serde_json::Error}       = "Serde conversion: {source}",
+    InvalidExpanded{message: String}       = "Expanded document invalid: {message}",
+    NoObject{document: Value}              = "Compacted document not a JSON object: {document}",
+}
+pub struct ExpandedJson(pub serde_json::Value);
```
ryanbtp, line 1173: This seems a bit unwieldy, is there no simple way to go from a context in json format to a ..context?

ryanbtp (approved):

mtbc:
Repushed as found a way to simplify the code (that I'd meant to try but forgot yesterday).

crates/common/src/prov/model/mod.rs
mtbc: If about `construct_context_definition`: I'd definitely expected one but I couldn't find it!

mtbc: Now wrapped it in a `lazy_static` at least though so we need do it no more than once.

ryanbtp (approved):

PR #239, opened 2023-03-23
replace PgEmbed with Testcontainers

mtbc:
Effects CHRON-268. There is some awkwardness in `make test` to move the `cargo test` outside `docker buildx` because the tests need to be able to create and access PostgreSQL containers.

Annoyingly I've not found mount flags that let me test with `--no-run` in building the test image then avoid the `docker exec` of `cargo test` doing much more downloading and building in the running container.


Makefile
```diff
@@ -78,8 +78,14 @@ tested-$(ISOLATION_ID): $(HOST_ARCHITECTURE)-ensure-context
 	rm -rf .artifacts
 	mkdir -p .artifacts
 
-	container_id=$$(docker create tested-artifacts:${ISOLATION_ID}); \
-		docker cp $$container_id:/artifacts `pwd`/.artifacts/ \
+	# "cargo test" itself uses Docker from within the container
```
ryanbtp, line 81: I have never tried this with buildx, but we should be able to docker-in-docker from cargo directly ?

mtbc: Not seeing any clear way to, alas.

scealiontach (changes requested):

Makefile
scealiontach: So you are putting everything together and then running the test outside of the image build?

mtbc: If we *can* get these tests running inside the image build as before then that would be great, it would probably solve my other to-do here of it running all the cargo downloading and compiling of libraries again even if I run a `cargo test` with `--no-run` inside the build first.

docker/unified-builder
```diff
@@ -148,13 +154,12 @@ RUN --mount=type=cache,target=/usr/local/cargo/registry,sharing=locked \
   echo "Unsupported architecture: $(BUILDPLATFORM)"; \
   exit 1; \
   fi \
-  && cargo test --locked --release --target ${CARGO_BUILD_TARGET}
+  && echo >run-tests cargo test --locked --release --target ${CARGO_BUILD_TARGET}
```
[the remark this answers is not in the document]
mtbc: It was about getting the correct `cargo test` command out to be runnable outside the build, as I didn't have `${CARGO_BUILD_TARGET}` available to the later `docker exec`.

Makefile
```diff
@@ -78,8 +78,14 @@ tested-$(ISOLATION_ID): $(HOST_ARCHITECTURE)-ensure-context
...
+	container_id=$$(docker run -d --network host \
+			-v /var/run/docker.sock:/var/run/docker.sock \
+			tested-artifacts:${ISOLATION_ID} sleep infinity); \
+		docker exec --user=root $$container_id chown root:docker /var/run/docker.sock \
```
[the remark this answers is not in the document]
mtbc: Didn't observe any knock-on effect to the host (which itself piqued my curiosity) but will definitely try out having `tester`'s UID match that of the user that runs make. :+1:

mtbc:
Here we now have a fourth commit that is closer to @scealiontach's thinking but for which the tests can't start their container so it fails CI. Investigation suggests there's no longer a `/var/run/docker.sock` at all, the question is of how to get the tests to use one.

Trying to keep the testing within `buildx`, and failing again with attempts at mounting, I've since experimented some with `socat` and docker networking to try to have that provide a route over TCP to the host's `docker.sock` but I've not yet devised a solution that works.

Is there a simple piece I'm still missing? My default plan is to persist a bit more with setting `DOCKER_HOST` in the build to enable it to reach the host's `docker.sock` even if I need `--allow-insecure-entitlement network.host` to achieve it.

mtbc:
Host networking turns out to be Linux-only so that won't let `make test` work on the Mac. Reaching the socket on the host via `socat`, I can indeed allow containerized tests to create containers, the problem then is that they can't connect to the exposed ports, which are listening on the host.

mtbc:
@ryanbtp reports vanilla `cargo test` troubles on Mac.

mtbc:
At @scealiontach's suggestion better to move back to running the tests from `docker run` rather than using `socat` at build time.

docker/unified-builder
```diff
@@ -126,29 +128,40 @@ RUN --mount=type=cache,target=/usr/local/cargo/registry,sharing=locked \
...
+RUN if [ "$BUILDPLATFORM" = "linux/arm64" ]; then \
   export CARGO_BUILD_TARGET=aarch64-unknown-linux-gnu; \
   elif [ "$BUILDPLATFORM" = "linux/amd64" ]; then \
   export CARGO_BUILD_TARGET=x86_64-unknown-linux-gnu; \
   else \
   echo "Unsupported architecture: $(BUILDPLATFORM)"; \
   exit 1; \
   fi \
-  && cargo test --locked --release --target ${CARGO_BUILD_TARGET}
+  && echo ${CARGO_BUILD_TARGET} >cargo-build-target
```
scealiontach, line 162: This is weird.  In the test, you aren´t going to have to do --target at all, since it will always be the buildplatform.

scealiontach (changes requested):

Makefile
```diff
@@ -73,8 +76,14 @@ tested-$(ISOLATION_ID): $(HOST_ARCHITECTURE)-ensure-context
 		--builder ctx-$(ISOLATION_ID)-$(HOST_ARCHITECTURE) \
 		--platform linux/$(HOST_ARCHITECTURE) \
 		--target test \
+		--build-arg DOCKER_GID=$(DOCKER_GID) \
+		--build-arg TESTER_UID=$(TESTER_UID) \
```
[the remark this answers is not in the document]
mtbc: How to run the tests under the UID of the user that runs make? Is there maybe a way to get the UID to the `useradd` without the build arg?

mtbc:
Experimentation suggests that the user running `cargo test` inside the container must be a member of a group that has the GID of `docker` on the host system.

mtbc:
I retained one `--build-arg DOCKER_GID=$(DOCKER_GID)` to arrange the [user being in the host's `docker`](https://github.com/btpworks/chronicle/pull/239#issuecomment-1490686611) (GID 132 on my laptop) but could achieve that with post-build surgery on the container if preferred.

docker/unified-builder
```diff
@@ -126,29 +128,40 @@ RUN --mount=type=cache,target=/usr/local/cargo/registry,sharing=locked \
...
+
+USER root
+RUN curl -fsSL https://download.docker.com/linux/debian/gpg | apt-key add - \
+  && echo "deb [arch=$(dpkg --print-architecture)] https://download.docker.com/linux/debian \
+  $(lsb_release -cs) stable" | tee /etc/apt/sources.list.d/docker.list > /dev/null \
+  && apt-get update -y \
+  && apt-get install -y --no-install-recommends \
+  docker-ce-cli
+
+RUN groupadd -g ${DOCKER_GID} docker \
```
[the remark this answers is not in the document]
mtbc: It does if I include the other packages but it is then added with a GID different from that on the host system which leaves me without permission to connect to Docker.

mtbc:
Now pushing another commit that illustrates that option. If indeed preferred, I'll squash it into the previous.

scealiontach (approved):

PR #243, opened 2023-04-04
introduce determinism checking into test infrastructure

suchapalaver:


mtbc (approved):
Looks good to me, maybe let's give @ryanbtp a chance to look tomorrow morning before merging as it's their ticket.

crates/opa-tp/src/tp.rs
```diff
@@ -453,26 +455,39 @@ mod test {
...
+                    match serde_json::from_str(&as_string) {
+                        Ok(json) => (k.clone(), json),
+                        Err(_) => (k.clone(), serde_json::to_value(v.clone()).unwrap()),
                     }
                 })
-                .collect()
+                .collect::<Vec<(String, Value)>>();
+
+            // Sort items by key to ensure determinism
+            items.sort_by_key(|(k, _)| k.to_owned());
```
ryanbtp, line 479: This is a BTree, so iteration should be in order anyway, what happened?

crates/opactl/src/main.rs
```diff
@@ -583,11 +622,127 @@ pub mod test {
...
+
+            // Log some debug information about the batch and sort its transactions.
+            debug!(received_batch = ?batch, transactions = ?batch.transactions);
+            let transactions = get_sorted_transactions(&mut batch);
+
+            // Perform a determinism checking cycle for a specified number of times
+            let number_of_determinism_checking_cycles = 5;
+
+            // Use the `test_determinism` function to perform determinism checking and get the expected contexts
+            let context = Arc::clone(&self.context);
```
ryanbtp, line 717: This is going to guarantee the sam context I think, as that arc contains a reference to the same context. We need a deep / non ref copy or a way of constructing a default one here. A constraint on context: default::Default might do it

ryanbtp, line 717: This is going to guarantee the sam context, as that arc contains a reference to the same context.

We need a deep / non ref copy or a way of constructing a default one here. A constraint on context: default::Default might do it - that way we can make new empty ones

ryanbtp (changes requested):

crates/opactl/src/main.rs
suchapalaver: If you have an arc: Arc<T> where T: Clone then you can do T::clone(&arc) to do a deep clone.

ryanbtp (approved):

PR #244, opened 2023-04-05
feat(charts): remove checked in dependencies

alexhq:
This undoes the temporary checked in dependencies added in https://github.com/btpworks/chronicle/pull/226


mtbc (approved):

alexhq:
Closed for now as we still require the checked in dependencies.

PR #246, opened 2023-04-06
report any error from starting poem endpoints

mtbc:
This fixes the issue that if `serve-graphql` can't bind to the socket then Chronicle would just exit without saying why.

It might be worth reviewing the other `.ok()`s in the startup code.


suchapalaver (approved):

mtbc:
Bah, server socket bind failure is being reported as a "File upload" error, I wonder how best to fix the usage of `thiserror` without breakage. At least it _is_ mentioning the  "Address already in use".

PR #249, opened 2023-04-10
add endpoint that offers JSON by IRI

mtbc:
With `serve-graphql` visit http://localhost:9982/data/chronicle:agent:my-agent-id and see JSON summarizing it. (Pipe through `jq .` for prettiness.) Same for activities and entities. Resolves CHRON-277 and CHRON-280.

Substituting `data` with `data/default` works, one may use other namespace IDs for `default`.

Notice that the subcommand has now become `serve-api` with a new `--offer-endpoints` option.

I opted for fairly minimal adjustment of `prov_model_for_namespace` code.

### DIFF
--- a/charts/chronicle/templates/statefulset.yaml
+++ b/charts/chronicle/templates/statefulset.yaml
@@ -72,7 +72,7 @@ spec:
               -c /etc/chronicle/config/config.toml \
               --console-logging json \
               --sawtooth tcp://{{ include "chronicle.sawtooth.service" . }}:{{ include "chronicle.sawtooth.sawcomp" . }} \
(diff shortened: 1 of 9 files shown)


ryanbtp:
Looks good but I missed a requirement here - for compact json ld we also need to have the context at an accessible url. So will need something like /context that serves our generated context, *and* the json-ld response will need to use a relative url to it rather than our current sort of symbolic one

mtbc:
Thank you, interesting, https://www.w3.org/TR/json-ld11/ confirms that,
> A remote context may also be referenced using a relative URL, which is resolved relative to the location of the document containing the reference.

ryanbtp (approved):

PR #253, opened 2023-04-12
fix(bootstrap,-sawtooth-tp): fix author

suchapalaver:

### DIFF
--- a/crates/chronicle/src/bootstrap/cli.rs
+++ b/crates/chronicle/src/bootstrap/cli.rs
@@ -912,7 +912,7 @@ impl SubCommand for CliModel {
         let mut app = Command::new("chronicle")
             .version(self.short_version.as_str())
             .long_version(self.long_version.as_str())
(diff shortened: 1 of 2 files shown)


crates/chronicle/src/bootstrap/cli.rs
```diff
@@ -912,7 +912,7 @@ impl SubCommand for CliModel {
         let mut app = Command::new("chronicle")
             .version(self.short_version.as_str())
             .long_version(self.long_version.as_str())
-            .author("Blockchain technology partners")
+            .author("BTPWorks")
```
mtbc, line 915:
```suggestion
            .author("BTP Works")
```
is how I've seen it to date but I may have missed a memo.

scealiontach: The legal entity here is still "Blockchain Technology Partners"  so it should stay the way it is.

PR #277, opened 2023-04-26
fix: Connectivity / startup robustness

ryanbtp:
* Retry loops (simple poll) for settings reading and OPA policy load
* Error level traces for timeout / loop
* Ensure we call reconnect on ZmgConnection after timeout or it can
  become unusable
* Similar retry loop on event subscription

### DIFF
--- a/crates/async-sawtooth-sdk/src/error.rs
+++ b/crates/async-sawtooth-sdk/src/error.rs
@@ -10,17 +10,20 @@ pub enum SawtoothCommunicationError {
     #[error("ZMQ error: {0}")]
     ZMQ(#[from] zmq::Error),
 
-    #[error("Send error {0}")]
-    Send(#[from] SendError),
+    #[error("Send error: {0}")]
+    SendChannel(#[from] SendError),
 
-    #[error("Receive error {0}")]
-    Receive(#[from] ReceiveError),
+    #[error("Sawtooth receive error: {0}")]
+    ReceiveSawtooth(#[from] ReceiveError),
 
-    #[error("Protobuf error {0}")]
+    #[error("Sawtooth send error: {0}")]
+    SendSawtooth(#[from] sawtooth_sdk::messaging::stream::SendError),
+
+    #[error("Protobuf error: {0}")]
     Protobuf(#[from] ProtobufError),
-    #[error("Protobuf decode error {0}")]
+    #[error("Protobuf decode error: {0}")]
     ProtobufProst(#[from] prost::DecodeError),
-    #[error("Unexpected Status {status:?}")]
+    #[error("Unexpected Status: {status:?}")]
     UnexpectedStatus { status: i32 },
     #[error("No transaction id for event")]
     MissingTransactionId,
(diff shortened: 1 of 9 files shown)


crates/async-sawtooth-sdk/src/error.rs
```diff
@@ -9,10 +11,13 @@ pub enum SawtoothCommunicationError {
     ZMQ(#[from] zmq::Error),
 
     #[error("Send error {0}")]
-    Send(#[from] SendError),
+    SendChannel(#[from] SendError),
+
+    #[error("Sawtooth receive error {0}")]
```
mtbc, line 16:
```suggestion
    #[error("Sawtooth receive error: {0}")]
```
Still keen on if we can start including colons or somesuch into all these for easier reading.

crates/async-sawtooth-sdk/src/zmq_client.rs
```diff
@@ -125,9 +120,9 @@ impl RequestResponseSawtoothChannel for ZmqRequestResponseSawtoothChannel {
                 .unwrap()
                 .send(message_type, &correlation_id, &bytes);
 
-            if let Err(e) = &future {
+            if let Err(e) = future {
                 error!(send_message=%correlation_id, error=%e);
-                //tx.send(Err(SawtoothCommunicationError::from(e.clone())));
+                tx.send(Err(SawtoothCommunicationError::from(e))).unwrap();
```
mtbc, line 125: The `unwrap()`'s a safe enough assumption? (Admittedly, already we do it again just below.)

crates/chronicle-protocol/src/settings.rs
```diff
@@ -0,0 +1,119 @@
...
+    /// Settings values are not uniform, so we return a `Vec<u8>` for further processing
+    pub async fn read_settings(&self, key: &str) -> Result<Setting, SawtoothCommunicationError> {
+        let address = sawtooth_settings_address(key);
+        loop {
+            let res = self.0.get_state_entry(&address).await;
+
+            if let Err(e) = res {
+                error!("Error reading settings: {}", e);
+                self.0.reconnect().await;
+                tokio::time::sleep(std::time::Duration::from_secs(2)).await;
```
mtbc, line 85: I wonder if we should make the increasing interval from `database.rs`'s connection retry something more general that anything can use. (I also wonder if it should be using `tokio::time` as here rather than `std::thread`.)

mtbc:
Any tips on how best to test this? Also an interesting challenge for CHRON-312 someday.

crates/async-sawtooth-sdk/src/zmq_client.rs
ryanbtp: Yes, although we could introduce an ordering where the channel can no longer send, this wont happen with the current code base

crates/chronicle-protocol/src/settings.rs
ryanbtp: std::thread is ok for the DB as there's no other IO going on at that point, generally no you wouldn't do this if you need the tokio reactor to work

ryanbtp:
Best (and currently only) way to test this is to start it without sawtooth (or with no opa / settings data), and to kill / restart the validator to see if things reconnect happily

crates/chronicle-protocol/src/settings.rs
ryanbtp: I can compose something on top of LedgerReader, but this is shippable, and it's fiddly to do in the general case until we replace the primatives in sawtooth-sdk as you have to recycle the blocking ZmqConnection or it breaks

mtbc: Thanks for giving the idea some thought and explaining. :smile:

mtbc (approved):

PR #278, opened 2023-04-27
have --interface option use ToSocketAddrs

mtbc:
Resolves CHRON-258.

Allows Chronicle to resolve `localhost` etc. as listening socket(s).

If any of the sockets cannot be bound, the server quits with an error, "Address already in use" or "Permission denied" or whichever is appropriate.


ryanbtp:
The second option is far preferable yes, going to be some overhead otherwise

ryanbtp (approved):

PR #284, opened 2023-05-02
fix: Respond to ping, improve logging

ryanbtp:
* Ping responder in recv_from_channel
* Non yielding polls now logged
* Other instrumentation improvements, including block id/height
* Compilation now set to panic-abort, so panicing background tasks will exit the process


suchapalaver (approved):

Makefile
```diff
@@ -13,6 +13,9 @@ HOST_ARCHITECTURE ?= $(shell uname -m | sed -e 's/x86_64/amd64/' -e 's/aarch64/a
 
 CLEAN_DIRS := $(CLEAN_DIRS)
 
+DOCKER_GID := $(shell getent group docker | cut -f 3 -d :)
```
mtbc, line 16: Don't need this but it's not a blocker.

crates/common/src/signing.rs
```diff
@@ -156,7 +156,9 @@ impl DirectoryStoredKeys {
         debug!(generate_secret_at = ?secretpath);
         let secret = SecretKey::random(StdRng::from_entropy());
 
-        let privpem = secret.to_pkcs8_pem(LineEnding::CRLF)?;
+        let privpem = secret
+            .to_pkcs8_pem(LineEnding::CRLF)
+            .map_err(|_| SignerError::InvalidPrivateKey {})?;
```
mtbc, line 161: Don't know if we might want to log these `_` but not a blocker for this PR.

ryanbtp: I think, it's a breaking change we did on main

PR #285, opened 2023-05-02
Fix/genesis

ryanbtp:
fix: Resolve genesis block
    
    * Searches from block 0, as unlike daml we can start with an empty
     stl


mtbc:
Certainly looks plausible.

PR #288, opened 2023-05-04
fix(bootstrap): fix import command handling of no input piped to stdin

suchapalaver:
Improve information around stdin not being readable, quit if if it's not and standard in is not a tty. 
[CHRON-323](https://blockchaintp.atlassian.net/browse/CHRON-323)

[CHRON-323]: https://blockchaintp.atlassian.net/browse/CHRON-323?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/crates/chronicle/src/bootstrap/mod.rs
+++ b/crates/chronicle/src/bootstrap/mod.rs
@@ -379,31 +379,29 @@ where
 
         Ok((ApiResponse::Unit, ret_api))
     } else if let Some(matches) = matches.subcommand_matches("import") {
-        let (namespace_id, namespace_uuid) = (
-            matches.value_of("namespace-id").unwrap(),
-            matches.value_of("namespace-uuid").unwrap(),
-        );
-        let namespace = NamespaceId::from_external_id(
-            namespace_id,
-            uuid::Uuid::try_parse(namespace_uuid)
-                .expect("cannot parse namespace UUID: {namespace_uuid}"),
-        );
+        let namespace = get_namespace(matches);
 
-        let contents = if matches.is_present("path") {
-            // Get the path to the file to import
-            let path: PathBuf = matches.value_of("path").unwrap().into();
-            let contents = std::fs::read_to_string(&path)?;
-            info!("Loaded import data from {:?}", path.display());
-            contents
+        let data = if let Some(path) = matches.value_of("path") {
+            let path = PathBuf::from(path);
+            let data = std::fs::read_to_string(&path)?;
+            info!("Loaded import data from {:?}", path);
+            data
         } else {
-            let mut buffer = String::new();
-            let mut stdin = io::stdin();
-            let bytes = stdin.read_to_string(&mut buffer)?;
-            info!("Loaded {bytes} bytes of import data from stdin");
-            buffer
+            if atty::is(atty::Stream::Stdin) {
+                eprintln!("Attempting to import data from standard input, press Ctrl-D to finish.");
+            }
+            info!("Attempting to read import data from stdin...");
+            let data = read_import_data()?;
+            info!("Loaded {} bytes of import data from stdin", data.len());
+            data
         };
 
-        let json_array = serde_json::from_str::<Vec<serde_json::Value>>(&contents)?;
+        if data.trim().is_empty() {
+            eprintln!("Import data is empty, nothing to import");
+            return Ok((ApiResponse::Unit, ret_api));
+        }
+
+        let json_array = serde_json::from_str::<Vec<serde_json::Value>>(&data)?;
 
         let mut operations = Vec::new();
         for value in json_array.into_iter() {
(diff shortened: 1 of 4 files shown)


crates/chronicle/src/bootstrap/mod.rs
```diff
@@ -396,13 +396,19 @@ where
             info!("Loaded import data from {:?}", path.display());
             contents
         } else {
+            eprintln!("Attempting to import data from stdin ... (Ctrl-D to finish)");
```
mtbc, line 399: :+1: This is exactly where `atty::is(Stream::Stdin)` might help so we don't print the message if we're piping from a file/stream.

crates/chronicle/src/bootstrap/mod.rs
```diff
@@ -396,13 +396,19 @@ where
...
             let mut buffer = String::new();
             let mut stdin = io::stdin();
             let bytes = stdin.read_to_string(&mut buffer)?;
             info!("Loaded {bytes} bytes of import data from stdin");
             buffer
         };
 
+        if contents.is_empty() {
```
mtbc, line 407: Maybe could also `trim()` in case was just a blank line or whatever.

crates/chronicle/src/bootstrap/mod.rs
suchapalaver: It's a cool idea. And certainly we do in further iterations, along with @ryanbtp's idea to chunk the input stream. We don't currently support importing data interactively, though.

crates/chronicle/src/bootstrap/mod.rs
suchapalaver: Nice idea, thanks.

crates/chronicle/src/bootstrap/mod.rs
suchapalaver: `atty::is(Stream::Stdin)` does not work for this purpose, but `grep_cli` does. See https://github.com/softprops/atty/issues/52#issuecomment-886751935.

suchapalaver: Okay, I think I've figured out what you meant now.

mtbc:
I was thinking of something more like,
```rust
let data = if let Some(path) = matches.value_of("path") {
    let path = PathBuf::from(path);
    let data = std::fs::read_to_string(&path)?;
    info!("Loaded import data from {:?}", path);
    data
} else {
    if atty::is(atty::Stream::Stdin) {
      eprintln!("Attempting to import data from standard input, press Ctrl-D to finish.");
    }
    info!("Attempting to read import data from stdin...");
    let data = read_import_data()?;
    info!("Loaded {} bytes of import data from stdin", data.len());
    data
};
```
so, for instance, `... import myns my-ns-uuid` gets the message about Ctrl-D, but `... import myns my-ns-uuid <from-my-data.json` does not get the message printed even though, in terms of CLAP arguments, they're  the same.

suchapalaver:
> I was thinking of something more like,
> 
> ```rust
> let data = if let Some(path) = matches.value_of("path") {
>     let path = PathBuf::from(path);
>     let data = std::fs::read_to_string(&path)?;
>     info!("Loaded import data from {:?}", path);
>     data
> } else {
>     if atty::is(atty::Stream::Stdin) {
>       eprintln!("Attempting to import data from standard input, press Ctrl-D to finish.");
>     }
>     info!("Attempting to read import data from stdin...");
>     let data = read_import_data()?;
>     info!("Loaded {} bytes of import data from stdin", data.len());
>     data
> };
> ```
> 
> so, for instance, `... import myns my-ns-uuid` gets the message about Ctrl-D, but `... import myns my-ns-uuid <from-my-data.json` does not get the message printed even though, in terms of CLAP arguments, they're the same.

Could you give what you're suggesting a try? I think you and `atty` might have differing definitions of what a tty is. Neither of the inputs you're giving as examples behave the way you're expecting.

mtbc:
I did try my suggestion before commenting. Have just done so again, still behaves exactly as I'd expect, so does your branch. If I give a filename or pipe something in, like from a file, I don't get the control+D message. If I leave it just attached to the terminal expecting input from that, I do get the message. Is that not what happens for you?

Maybe we can screenshare on Monday so I can see what you're seeing.

suchapalaver:
> I did try my suggestion before commenting. Have just done so again, still behaves exactly as I'd expect, so does your branch. If I give a filename or pipe something in, like from a file, I don't get the control+D message. If I leave it just attached to the terminal expecting input from that, I do get the message. Is that not what happens for you?
> 
> 
> 
> Maybe we can screenshare on Monday so I can see what you're seeing.

Could you share the exact command you used to get the Ctrl + D message path, please?

mtbc:
```
mtbc@pop-os:~/src/chronicle$ git checkout origin/fix/what-is-a-tty
HEAD is now at 44a2049 fix(bootstrap): fix import command hanging if no input piped to stdin
mtbc@pop-os:~/src/chronicle$ git status
HEAD detached from origin/main
nothing to commit, working tree clean
mtbc@pop-os:~/src/chronicle$ cargo run --bin chronicle --features inmem -- import testns 60976119-adc6-3826-a57b-3ccb0580e7bd 
    Finished dev [unoptimized + debuginfo] target(s) in 0.18s
     Running `target/debug/chronicle import testns 60976119-adc6-3826-a57b-3ccb0580e7bd`
Attempting to import data from standard input, press Ctrl-D to finish.
```

suchapalaver:
> I did try my suggestion before commenting. Have just done so again, still behaves exactly as I'd expect, so does your branch. If I give a filename or pipe something in, like from a file, I don't get the control+D message. If I leave it just attached to the terminal expecting input from that, I do get the message. Is that not what happens for you?
> 
> Maybe we can screenshare on Monday so I can see what you're seeing.

Sorry! You were right! Don't know why I wasn't seeing it before. I made all your suggested changes.

mtbc (approved):

PR #290, opened 2023-05-04
Fix/panic hook

ryanbtp:
* Abort with a decent stacktrace on thread panics
* Write-retry on error
* Use updated sawtooth-sdk-rust that doesn't panic when a receiver has dropped

### DIFF
--- a/crates/chronicle/src/bootstrap/mod.rs
+++ b/crates/chronicle/src/bootstrap/mod.rs
@@ -29,7 +29,7 @@ use sawtooth_protocol::{events::StateDelta, messaging::SawtoothSubmitter};
 use telemetry::{self, ConsoleLogging};
 use url::Url;
 
-use std::{io, net::SocketAddr};
+use std::{backtrace::Backtrace, io, net::SocketAddr, panic};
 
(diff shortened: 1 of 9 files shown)


crates/chronicle/src/bootstrap/mod.rs
```diff
@@ -333,6 +333,14 @@ pub async fn bootstrap<Query, Mutation>(
     Query: ObjectType + 'static + Copy,
     Mutation: ObjectType + 'static + Copy,
 {
+    //panic::always_abort();
```
suchapalaver, line 336:
```suggestion
```

crates/chronicle/src/bootstrap/mod.rs
```diff
@@ -333,6 +333,14 @@ pub async fn bootstrap<Query, Mutation>(
...
+    panic::set_hook(Box::new(|info| {
+        //let stacktrace = Backtrace::capture();
```
suchapalaver, line 338:
```suggestion
```

suchapalaver (approved):

suchapalaver (approved):

PR #293, opened 2023-05-08
fix/port 06 comms fixes

ryanbtp:
* Ping response that doesn't drive stream events to consumers
* Non panicing sawtooth zmq
* Refactored blockid / num for clarity
* Better distringuish start points as id, genesis or head
* Model marker blockids rather than use option
* Compose send-reconnect-retry with trait
* Improved last seen block implementation
* Copy types for position / blockid
* Introduce retry in opactl
* Consistent version number in log and help text for all executables, including git hash

Ref: CHRON-345 CHRON-328 CHRON-231


mtbc:
Looks good to me but Jenkins not yet pleased.

.VERSION
```diff
@@ -0,0 +1 @@
+v0.6.0-311-gf435e33b
```
mtbc, line 1: Is this something we should commit?

crates/async-sawtooth-sdk/src/ledger.rs
```diff
@@ -24,19 +31,53 @@ use sawtooth_sdk::messages::{
...
+    fn try_from(s: String) -> Result<BlockId, Infallible> {
+        Ok(hex::decode(s)
+            .map_err(BlockIdError::from)
+            .and_then(|x| {
+                x.as_slice()
+                    .try_into()
+                    .map_err(BlockIdError::from)
+                    .map(BlockId::Block)
+            })
+            .unwrap_or(BlockId::Marker))
```
mtbc, line 71: Should we check they really are zeroes?

.VERSION
ryanbtp: It should probably be 'local', but as it's a compile time resource it's fiddly to default in a macro. Has to be done externally to cargo, and I want to keep cargo build (not just make) working out of the box as much as possible

crates/async-sawtooth-sdk/src/ledger.rs
ryanbtp: The binary protocol using hex encoded strings for hashes in the first place leads down this dark path :)

ryanbtp: They are case sensitive too! I found that out the hard way

suchapalaver (approved):

PR #294, opened 2023-05-09
adjust message sending on to zmq

mtbc:
* expands logging
* retries `send` after `TimeoutError`
* always uses new correlation ID

### DIFF
--- a/crates/sawtooth-protocol/src/messaging.rs
+++ b/crates/sawtooth-protocol/src/messaging.rs
@@ -140,24 +140,31 @@ impl SawtoothSubmitter {
                 batches: vec![batch],
             };
 
+            let correlation_id = uuid::Uuid::new_v4().to_string();
+            debug!("submitting request with correlation ID {correlation_id}");
+
             let mut future = loop {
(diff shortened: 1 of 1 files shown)


crates/sawtooth-protocol/src/messaging.rs
```diff
@@ -140,24 +140,30 @@ impl SawtoothSubmitter {
...
 
-                // Force reconnection on any send error that's not a timeout -
-                // disconnect can actually mean a dead Zmq Thread
-                if let Err(SendError::UnknownError) | Err(SendError::DisconnectedError) = future {
-                    debug!("Send error, re-initialise ZMQ");
-                    let (tx, rx) = ZmqMessageConnection::new(self.address.as_str()).create();
-                    self.tx = tx;
-                    self.rx = rx;
-                    continue;
+                if let Err(send_error) = &future {
```
scealiontach, line 153: I'd put a log before this line too.  The later messages won't be issued if it is hanging on the future itself.

mtbc:
Superseded by #296.

PR #308, opened 2023-05-23
build authentication server image

mtbc:
For the JWKS-enabled Helm testing in CHRON-330 we need a source of verifiable authorization tokens. This PR isn't a proposal for how to include this with Chronicle's testing, it's more a current means to provide the authentication server side for the testing.

Notes on actually using this with `curl` and Chronicle:

1. While the container is running (e.g., during a part of `make test` for this PR) one may `docker exec <container> oauth-token` to obtain a new bearer token arbitrarily many times.
2. Add this token to API requests with some equivalent of curl's `-H "Authorization: Bearer <token>"` filling in the provided token.
3. Forward a local port in Chronicle's container into this new container. That is, Chronicle's `http://localhost:8090/` should reach this container's exposed TCP 8090. (I didn't string that together for this PR because I figure we're using helm, not docker compose, anyway.)
4. Chronicle's settings may then include `--require-auth --id-claims email --jwks-address http://localhost:8090/jwks --userinfo-address http://localhost:8090/userinfo` where we are using the `email` field to make doubly sure that the `/userinfo` endpoint is being used. If you try your oauth-token at https://jwt.io/ then you will see it doesn't have `email` already.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -8,3 +8,4 @@
 !docker/chronicle-config
 !docker/chronicle-test
 !docker/chronicle-helm-test
+!docker/id-provider-settings.yml
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2257,6 +2257,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "hyper-rustls"
+version = "0.23.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1788965e61b367cd03a62950836d5cd41560c3577d90e40e0819373194d1661c"
+dependencies = [
+ "http",
+ "hyper",
+ "rustls",
+ "tokio",
+ "tokio-rustls",
+]
+
+[[package]]
 name = "hyper-tls"
 version = "0.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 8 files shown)


mtbc:
The `wait-for-it` complaints from CI aren't much relevant.

suchapalaver (approved):

suchapalaver:
The image wasn't built for me when I ran `make build`.

suchapalaver:
I accidentally hit the close button earlier. Apologies!

mtbc:
@suchapalaver : Thank you for the review. Is it enabling for you if I fix the CI issue then merge this, as a basis for what you're doing? (I could just omit the wait-for-it part to keep my touch light.)

suchapalaver:
> @suchapalaver: Thank you for the review. Is it enabling for you if I fix the CI issue then merge this, as a basis for what you're doing? (I could just omit the wait-for-it part, or even the entire second commit, to keep my touch light, or you're welcome to just cherry-pick the first commit.)

Me cherry-picking is just to keep work moving. My sense is that this *should* be a published image since it may be used in long-running testing scenarios.

mtbc:
The changes to the id-provider image in the latest commits match the change in #310.

mtbc:
(Dropped the `wait-for-it` complication after testing locally.)

mtbc:
Needs re-review given the changes.

ryanbtp (approved):

PR #304, opened 2023-05-17
fix: update helm chart Chronicle version

suchapalaver:
[CHRON-351](https://blockchaintp.atlassian.net/browse/CHRON-351)

PR #309, opened 2023-05-24
fix: Stream completion for opactl / async-sawtooth-sdk

ryanbtp:
* Ensure Shutdown is sent to ZmqMessageConnection before shutting down tasks to avoid panic
* Use select! and join! to simplify task waiting
* Calculate number of blocks remaining before timeout correctly, and log it
  in a readable way
* Log stream completion, no need to handle none with the select!
* Various instrumentation improvements
* Blockid width, warn on parse failiures
* Tested with high concurrency to induce overlapped transaction writes /
  waits in multiple clients

Ref: CHRON-358

# PR Checklist

### DIFF
--- a/crates/api/src/lib.rs
+++ b/crates/api/src/lib.rs
@@ -282,7 +282,7 @@ where
         let start_from_block = if let Ok(Some(start_from_block)) = last_seen_block {
             FromBlock::BlockId(start_from_block)
         } else {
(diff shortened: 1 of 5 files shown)


suchapalaver:
Cherry-picked this into [304](https://github.com/btpworks/chronicle/pull/304) and the bootstrap process for `opactl` is not returning an error.

suchapalaver (approved):

PR #312, opened 2023-05-28
URL OPA Policy Loader

suchapalaver:
- [ ] `common` library `opa` library support for a `UrlPolicyLoader` - [task](https://blockchaintp.atlassian.net/browse/CHRON-363)
    - tests
        - [ ] mock http opa bundle server
        - [ ] bare path loading
        - [ ] file URL - pain to mock because of the URL type format
- [ ] CLI option to  `--load-opa-from-url` [task](https://blockchaintp.atlassian.net/browse/CHRON-364):

```text
USAGE:
    chronicle serve-api [OPTIONS] --load-opa-from-url <load-opa-from-url> <--opa-policy-name <opa-policy-name>|--opa-policy-entrypoint <opa-policy-entrypoint>>
    --load-opa-from-url <load-opa-from-url>
            URL for loading OPA policy bundle
    --opa-policy-entrypoint <opa-policy-entrypoint>
            Entrypoint to the OPA policy if 'load-opa-from-url' is selected
    --opa-policy-name <opa-policy-name>
            Name of the OPA policy to use if 'load-opa-from-url' is selected
```

[CHRON-361](https://blockchaintp.atlassian.net/browse/CHRON-361)

# PR Checklist


crates/chronicle/src/bootstrap/cli.rs
```diff
@@ -1056,6 +1056,33 @@ impl SubCommand for CliModel {
                         .value_parser(["data", "graphql"])
                         .default_values(&["data", "graphql"])
                         .help("which API endpoints to offer")
+                    )
+                    .arg(
+                        Arg::new("load-opa-from-url")
```
mtbc, line 1061:
```suggestion
                        Arg::new("opa-policy-address")
```
Just a suggestion, not a blocker. (Thinking that the prefix fits the other OPA policy stuff and the suffix fits other URL stuff among our options.) Curious about other ideas though given that "policy address" has a non-tech meaning!

People might appreciate an environment variable for these too.

crates/chronicle/src/bootstrap/cli.rs
```diff
@@ -1056,6 +1056,33 @@ impl SubCommand for CliModel {
...
+                        Arg::with_name("opa-policy-name")
+                            .long("opa-policy-name")
+                            .help("Name of the OPA policy to use if 'load-opa-from-url' is selected")
+                            .takes_value(true)
+                            .requires("load-opa-from-url"),
+                    )
+                    .arg(
+                        Arg::with_name("opa-policy-entrypoint")
+                            .long("opa-policy-entrypoint")
+                            .help("Entrypoint to the OPA policy if 'load-opa-from-url' is selected")
```
mtbc, line 1078:
```suggestion
                            .help("Entrypoint to the OPA policy")
```
The latter part soon becomes clear anyway given the requires so can simplify. Again, just a suggestion.

crates/chronicle/src/bootstrap/cli.rs
```diff
@@ -1056,6 +1056,33 @@ impl SubCommand for CliModel {
...
+                            .takes_value(true)
+                            .requires("load-opa-from-url"),
+                    )
+                    .group(
+                        ArgGroup::with_name("opa-policy-from-url-args")
+                            .args(&["opa-policy-name", "opa-policy-entrypoint"])
+                            .requires("load-opa-from-url"),
```
mtbc, line 1085: Curious what this adds given that the arguments separately also require it.

crates/chronicle/src/bootstrap/mod.rs
```diff
@@ -278,6 +281,21 @@ async fn configure_opa(config: &Config, options: &ArgMatches) -> Result<Configur
             "Chronicle operating in an insecure mode with an embedded default OPA policy"
         );
         Ok(ConfiguredOpa::Embedded(opa))
+    } else if options.is_present("opa-policy-from-url") {
```
mtbc, line 284: `load-opa-from-url`?

crates/chronicle/src/bootstrap/cli.rs
```diff
@@ -1056,6 +1056,33 @@ impl SubCommand for CliModel {
                         .value_parser(["data", "graphql"])
                         .default_values(&["data", "graphql"])
                         .help("which API endpoints to offer")
+                    )
+                    .arg(
```
mtbc, line 1060: Should these OPA options be on the `serve-api` subcommand only? Others like `import` may also need them.

crates/common/src/opa.rs
```diff
@@ -280,6 +292,104 @@ impl PolicyLoader for CliPolicyLoader {
...
+    }
+
+    fn load_policy_from_bytes(&mut self, policy: &[u8]) {
+        self.policy = policy.to_vec();
+    }
+
+    async fn load_policy(&mut self) -> Result<(), PolicyLoaderError> {
+        let url = &self.address;
+
+        let bundle = if url.starts_with("http://")
```
mtbc, line 361: No need to repeat parsing logic when we already have it. (Also, what of "HTTP://" etc.?) Just see if `Url::parse` likes it or not and, if it doesn't, fall back to parsing as `Path`?

mtbc:
Interesting to see your use of a mockito HTTP server, could be a useful touch in other tests, nice find.

crates/common/src/opa.rs
```diff
@@ -280,6 +292,104 @@ impl PolicyLoader for CliPolicyLoader {
...
+            || url.starts_with("https://")
+            || url.starts_with("file://")
+        {
+            self.fetch_bundle_from_url().await?
```
mtbc, line 365: Probably good to test that `reqwest` even works for file URLs.

suchapalaver: But how to mock a file URL was what I couldn't figure out. Is there a suitable target we can test with?

mtbc: `test_load_policy_from_bare_path` does a fine job of creating a temporary file, I'd think it should work to take a similar approach but prefix its absolute path with `file://` to construct a good test target.

crates/chronicle/src/bootstrap/cli.rs
suchapalaver: Think I've fixed this. Thanks

crates/common/src/opa.rs
suchapalaver: Can we do another round of review on this bit, please? Not there yet, I don't think, but a bit closer.

crates/common/src/opa.rs
suchapalaver: Ah, I'll give it a go. Confused by thinking a file URL refers to a file that lives online somewhere.

suchapalaver:
@mtbc Will do a final round on the documentation once we've nailed down everything else.

crates/common/src/opa.rs
mtbc: Sure thing, will take another look today.

crates/chronicle/src/bootstrap/cli.rs
```diff
@@ -982,6 +982,31 @@ impl SubCommand for CliModel {
                     .help("Name of the database")
                     .default_value("chronicle"),
             )
+            .arg(
+                Arg::new("opa-bundle-address")
+                .long("opa-bundle-address")
+                .required(false)
```
mtbc, line 988: Isn't this the default?

crates/chronicle/src/bootstrap/cli.rs
```diff
@@ -982,6 +982,31 @@ impl SubCommand for CliModel {
...
+            .arg(
+                Arg::with_name("opa-policy-name")
+                    .long("opa-policy-name")
+                    .help("Name of the OPA policy to be used")
+                    .takes_value(true)
+            )
+            .arg(
+                Arg::with_name("opa-policy-entrypoint")
+                    .long("opa-policy-entrypoint")
+                    .help("Entrypoint to the intended OPA policy")
```
mtbc, line 1002:
```suggestion
                    .help("Entrypoint to the named OPA policy")
```

crates/common/src/opa.rs
```diff
@@ -30,6 +36,12 @@ pub enum PolicyLoaderError {
 
     #[error("Error loading OPA policy: {0}")]
     SawtoothCommunicationError(#[from] SawtoothCommunicationError),
+
+    #[error("URL request error: {0}")]
+    UrlRequestError(#[from] reqwest::Error),
+
+    #[error("URL IO error: {0}")]
+    UrlIoError(#[from] std::io::Error),
```
mtbc, line 44: No objection but it's worth being aware that if someone adds further code to this crate using suffix `?` for some *other* source of IO errors then it may surprise them to find it being reported as being URL-related.

crates/common/src/opa.rs
```diff
@@ -456,4 +568,126 @@ mod tests {
         }
         Ok(())
     }
+
+    fn embedded_policy_bundle() -> Result<Vec<u8>, PolicyLoaderError> {
+        let bundle_file = "bundle.tar.gz";
```
mtbc, line 573: Could be a `const`?

mtbc:
Looking good overall, will test some more tomorrow. :+1:

crates/common/src/opa.rs
```diff
@@ -280,6 +292,106 @@ impl PolicyLoader for CliPolicyLoader {
...
+                let path = Path::new(address);
+                if path.is_file() {
+                    std::fs::read(path)?
+                } else {
+                    return Err(PolicyLoaderError::InvalidPath(address.to_string()));
+                }
+            }
+        };
+
+        info!(loaded_policy_bytes=?bundle.len(), "Loaded policy");
```
mtbc, line 378: Seems odd to log "loaded policy" before calling `load_policy_from_bundle`.
```suggestion
        info!(fetched_policy_bytes=?bundle.len(), "Fetched policy");
```

docs/cli.md
```diff
@@ -123,3 +123,47 @@ chronicle \
 
 Chronicle will also generate subcommands for recording provenance, derived from
 your [domain configuration](./domain_modeling.md).
+
+## Load OPA Policy from URL Arguments
+
+### `--opa-bundle-address <--opa-policy-name |--opa-policy-entrypoint`
```
mtbc, line 129: Docs still need updating before we merge.

docs/opa.md
```diff
@@ -479,3 +479,52 @@ chronicle.opa.entrypoint=allow_transactions.allowed_users
...
+For example:
+
+```text
+--opa-policy-name my_policy --opa-policy-entrypoint entrypoint1
+```
+
+Note that when using `--opa-bundle-address` option, the `--opa-policy-name` and
+`--opa-policy-entrypoint` options must be provided.
+
+If you prefer to load the OPA policy bundle directly from a file path without using
```
mtbc, line 520: We seem to pretty much say this end part above where HTTP and file path get equal billing.

mtbc:
CLI side definitely looking improved, thank you. :+1:

docs/cli.md
```diff
@@ -123,3 +123,47 @@ chronicle \
 
 Chronicle will also generate subcommands for recording provenance, derived from
 your [domain configuration](./domain_modeling.md).
+
+## Load OPA Policy from URL or File Arguments
+
+### `--opa-bundle-address --opa-policy-name --opa-policy-entrypoint`
```
mtbc, line 129: Perhaps include the arguments in angle brackets to fit with the other options this page describes.

docs/opa.md
suchapalaver: Okay. Done.

docs/cli.md
```diff
@@ -123,3 +123,51 @@ chronicle \
...
+
+#### `--opa-policy-name <name>`
+
+The `--opa-policy-name` option specifies the name of the OPA policy.
+
+#### `--opa-policy-entrypoint <entrypoint>`
+
+The `--opa-policy-entrypoint` option specifies the entrypoint within the policy.
+
+Note that `--opa-bundle-address` requires both the `--opa-policy-name` and
```
mtbc, line 143: Maybe move this up to the section about the address.

mtbc (approved):

PR #314, opened 2023-05-30
add logging to opactl

mtbc:
Resolves CHRON-362. Adding some log messages seems to have enabled various more from spans.

# PR Checklist

### DIFF
--- a/crates/opactl/src/cli.rs
+++ b/crates/opactl/src/cli.rs
@@ -8,7 +8,7 @@ use clap::{
 use k256::{pkcs8::DecodePrivateKey, SecretKey};
 use rand::rngs::StdRng;
 use rand_core::SeedableRng;
-use tracing::{error, info};
+use tracing::{debug, error, info};
 use url::Url;
 
 // Generate an ephemeral key if no key is provided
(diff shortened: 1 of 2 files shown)


crates/opactl/src/cli.rs
```diff
@@ -311,25 +311,28 @@ pub fn cli() -> Command {
...
-                error!("Unable to read file {path}");
-                exit(1)
-            });
-            SecretKey::from_pkcs8_pem(key.trim()).unwrap()
-        }
-        ValueSource::EnvVariable => {
-            let key: &String = matches.get_one(name).unwrap();
-            SecretKey::from_pkcs8_pem(key.trim()).unwrap()
+    if let Some(source) = matches.value_source(name) {
+        debug!("loading key \"{name}\" from {source:?}");
```
ryanbtp, line 315: These macros are probably better written to exploit structured logging rather than generating strings.

```
debug!(loading_key=?name, from=?source)
```

ryanbtp (changes requested):

crates/opactl/src/cli.rs
ryanbtp: Certainly by the time it ends up in a modern log aggregator like Loki, having fields already split out into k/v really reduces the amount of time your analytics spend running regular expressions

ryanbtp (approved):

PR #322, opened 2023-06-05
feat: Load policy from remote or path

ryanbtp:
Ref: CHRON-337

# PR Checklist

### DIFF
--- a/crates/opactl/src/main.rs
+++ b/crates/opactl/src/main.rs
@@ -35,7 +35,7 @@ use rand::rngs::StdRng;
 use rand_core::SeedableRng;
 use tokio::runtime::Handle;
 use tracing::{debug, error, info, instrument, span, trace, Level};
-use url::Url;
+use url::{ParseError, Url};
 use user_error::UFE;
 mod cli;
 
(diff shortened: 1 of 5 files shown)


crates/opactl/src/main.rs
```diff
@@ -312,13 +315,13 @@ async fn dispatch_args<
         Some(("set-policy", matches)) => {
             let root_key: SigningKey = load_key_from_match("root-key", matches).into();
             let transactor_key: SigningKey = load_key_from_match("transactor-key", matches).into();
-            let policy: &PathBuf = matches.get_one("policy").unwrap();
-            let mut policy_file = File::open(policy)?;
-            let mut policy = vec![];
-            policy_file.read_to_end(&mut policy)?;
+            let policy: &String = matches.get_one("policy").unwrap();
```
suchapalaver, line 318: We can currently get the policy into the Helm Chart via the built default policies that are built into `opactl`. Can we not keep that option and have another `--from-url` option? Otherwise, without a policy bundle at an address, this breaks the set-policy support in the Helm Chart.

suchapalaver (changes requested):

crates/opactl/src/main.rs
scealiontach: to @suchapalaver's point, and the title of the PR,  this should also take a file path, and `reqwest` apparently does not, nor even a `file://`  url

ryanbtp: I'll stick a protocol switch in front

suchapalaver: Building and testing with the chart.

crates/opactl/src/main.rs
```diff
@@ -312,10 +321,28 @@ async fn dispatch_args<
...
+
+            enum UrlOrFile {
+                Url(Url),
+                File(PathBuf),
+            }
+
+            let policy = match policy.parse::<Url>() {
```
suchapalaver, line 331: I'm pretty sure `file://bundle.tar.gz` type urls will parse successfully here.

suchapalaver (changes requested):

crates/opactl/src/main.rs
ryanbtp: And they are absolute too, resolved and documented

suchapalaver (approved):

PR #340, opened 2023-06-08
fix(charts): fix auth args handling

suchapalaver:
Fixes an issue with the current chart where it's assumed that `--require-auth` is required in order to enable Chronicle's auth support features.

---
## Scenarios

The Helm Chart installs a `devIdProvider` by default, which provides a token to the api test. `devIdProvider.enabled: true` means that if no other `auth` values are provided - `auth.id.claims`, `auth.jwks.url`, `auth.userinfo.url` - Chronicle and its testing suite will be provided the jwks and userinfo urls for the `devIdProvider`.
 
```yaml
devIdProvider:
  enabled: true
```

`auth.required` provides the equivalent of the `--auth-required` CLI flag. It requires `auth.jwks.url`; by default it will fall back to the jwks url of the `devIdProvider`. If `auth.id.claims` are not provided Chronicle will default to `[iss, sub]`. If they are provided, those will be used by Chronicle.

```yaml
auth:
  required: false
  id:
    claims:
  jwks:
    url:
  userinfo:
    url:
```

If `devIdProvider.enabled: true` and the user provides either or both `auth.jwks.url` and `auth.userinfo.url` the installation will provide these values to Chronicle. 

Whether a Chronicle installation supports JWT identities depends on whether the `devIdProvider` is enabled, and whether the user provides either the `auth.jwks.url` or `auth.userinfo.url`. Whether Chronicle once installed requires users to be authenticated via JWT depends on whether `auth.required` is set to `true` or `false`.

In testing, by default `test.enabled` is set to `true`, which enables the test suite. If `auth.required: true` when testing, the `devIdProvider should also be enabled` so that the api-test includes an authorization header with the mutation it sends to Chronicle's GraphQL API.

```yaml
test:
  enabled: true
``` 

```yaml
auth:
  required: true
```

[CHRON-397](https://blockchaintp.atlassian.net/browse/CHRON-397)

## To Test ...

```yaml
---
imagePullSecrets:
  enabled: true
  value: 
    - name: regcred

auth:
  required: true

opa:
  enabled: true

test:
  api:
    image:
      pullPolicy: Always
      repository: dev.catenasys.com:8083/chronicle-helm-api-test-amd64
      tag: latest
  enabled: true
```

# PR Checklist

[CHRON-397]: https://blockchaintp.atlassian.net/browse/CHRON-397?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/charts/chronicle/templates/_chronicle.tpl
+++ b/charts/chronicle/templates/_chronicle.tpl
@@ -55,20 +55,64 @@ chronicle: {{ include "common.names.fullname" . }}
 {{ include "common.names.fullname" . }}-test-id-provider
 {{- end -}}
 
-{{- define "chronicle.jwksUrl" -}}
+{{- define "chronicle.id-provider.service.url" -}}
+http://{{ include "chronicle.id-provider.service" . }}:8090/jwks
+{{- end -}}
+
+{{- define "chronicle.id-claims" -}}
+{{- if .Values.auth.id.claims -}}
+--id-claims {{ .Values.auth.id.claims }} \
+{{- else -}}
+{{- /* Do nothing */ -}}
+{{- end -}}
+{{- end -}}
+
+{{/* The JWKS and userinfo URLs are connected. */}}
+{{/* If either is provided Chronicle will use the user-provided options. */}}
+{{/* If neither is provided Chronicle should fall back to using the 'devIdProvider'.*/}}
+{{- define "chronicle.jwks-url.url" -}}
+{{- if or (.Values.auth.jwks.url) (.Values.auth.userinfo.url) -}}
 {{- if .Values.auth.jwks.url -}}
 {{ .Values.auth.jwks.url }}
+{{- end -}}
 {{- else -}}
 {{- if .Values.devIdProvider.enabled -}}
-http://{{ include "chronicle.id-provider.service" . }}:8090/jwks
+{{ include "chronicle.id-provider.service.url" . }}
 {{- else -}}
-{{ required "devIdProvider.enabled must be true or auth.jwks.url must be set!" .Values.auth.jwks.url }}
+{{/* Do nothing */}}
 {{- end -}}
 {{- end -}}
 {{- end -}}
 
-{{- define "chronicle.userinfoUrl" -}}
+{{- define "chronicle.jwks-url.cli" -}}
+{{- if or (.Values.auth.jwks.url) (.Values.devIdProvider.enabled) -}}
+--jwks-address {{ include "chronicle.jwks-url.url" . }} \
+{{- end -}}
+{{/* Do nothing */}}
+{{- end -}}
+
+{{/* The JWKS and userinfo URLs are connected. */}}
+{{/* If either is provided Chronicle will use the user-provided options. */}}
+{{/* If neither is provided Chronicle should fall back to using the 'devIdProvider'.*/}}
+{{- define "chronicle.userinfo-url" -}}
+{{- if or (.Values.auth.jwks.url) (.Values.auth.userinfo.url) -}}
+{{- if .Values.auth.userinfo.url -}}
 {{ .Values.auth.userinfo.url }}
+{{- end -}}
+{{- else -}}
+{{- if .Values.devIdProvider.enabled -}}
+{{ include "chronicle.id-provider.service.url" . }}
+{{- else -}}
+{{/* Do nothing */}}
+{{- end -}}
+{{- end -}}
+{{- end -}}
+
+{{- define "chronicle.userinfo-url.cli" -}}
+{{- if or (.Values.auth.userinfo.url) (.Values.devIdProvider.enabled) -}}
+--userinfo-address {{ include "chronicle.userinfo-url" . }} \
+{{- end -}}
+{{/* Do nothing */}}
 {{- end -}}
 
 {{- define "chronicle.root-key.secret" -}}
--- a/charts/chronicle/templates/statefulset.yaml
+++ b/charts/chronicle/templates/statefulset.yaml
@@ -104,7 +104,7 @@ spec:
             - |
               {{ if .Values.auth.required }}
               {{ if and (not .Values.auth.jwks.url) (not .Values.auth.userinfo.url) (not .Values.devIdProvider.enabled) }}
-              {{ required "If auth.required you need to provide at least auth.jwks.url or auth.userinfo.url" .Values.auth.jwks.url }}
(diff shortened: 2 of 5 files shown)


charts/chronicle/templates/_chronicle.tpl
```diff
@@ -68,8 +68,12 @@ http://{{ include "chronicle.id-provider.service" . }}:8090/jwks
 {{- end -}}
 
 {{- define "chronicle.userinfoUrl" -}}
+{{- if .Values.devIdProvider.enabled -}}
```
scealiontach, line 71: Not good enough - you need the similar logic as for `chronicle.jwksUrl`  It's legitimate to keep the devIdProvider around, but use a different URL.

charts/chronicle/templates/statefulset.yaml
```diff
@@ -125,11 +125,17 @@ spec:
                   --interface 0.0.0.0:{{ .Values.port}} \
                   {{- if .Values.auth.required }}
                   --require-auth \
+                  {{- end }}
+                  {{- if .Values.auth.id.claims }}
```
scealiontach, line 129: you have a default value for this  so it is always going to be true unless the user explicitly sets it to nil

charts/chronicle/templates/statefulset.yaml
```diff
@@ -125,11 +125,17 @@ spec:
...
                   --id-claims {{ .Values.auth.id.claims }} \
+                  {{- end }}
                   {{- if .Values.auth.jwks.enabled }}
                   --jwks-address {{ include "chronicle.jwksUrl" . }} \
                   {{- end }}
-                  {{- if .Values.auth.userinfo.url }}
+                  {{- if and (not .Values.devIdProvider.enabled) (not .Values.auth.jwks.enabled) }}
```
scealiontach, line 135: shouldn't this now be `or .Values.devIdProvider.enabled .Values.jwks.url`

scealiontach (changes requested):

charts/chronicle/templates/_chronicle.tpl
suchapalaver: The changes didn't "outdate" your comment, but I think it's been addressed.

docs/helm-jwks-opa.md
```diff
@@ -183,6 +184,7 @@ disable OIDC:
 
 ```yaml
 auth:
+  enabled: true
```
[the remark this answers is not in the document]
suchapalaver: I think I've found a way to get rid of it. Just providing the `test.auth.enabled` option for testing now.

charts/chronicle/templates/statefulset.yaml
suchapalaver: Default is gone. Chronicle provides `[iss, sub]` by default.

mtbc:
`helm-jwks-opa.md` should also now mention the behavior where `required: true` is omitted yet an OIDC endpoint URI is set. If preferred, I can help with this in a subsequent PR.

suchapalaver:
> `helm-jwks-opa.md` should also now mention the behavior where `required: true` is omitted yet an OIDC endpoint URI is set. If preferred, I can help with this in a subsequent PR.

Since documentation can be published on a rolling basis i would be great to do a subsequent roundup on that once this is done, that'd be great.

charts/chronicle/templates/_chronicle.tpl
```diff
@@ -52,23 +52,70 @@ chronicle: {{ include "common.names.fullname" . }}
 {{- end -}}
 
 {{- define "chronicle.id-provider.service" -}}
+{{- if .Values.IdProvider.service -}}
```
scealiontach, line 55: I dont't see the point of this value.  The only place you seem to be using this template is when the devIdProvider.enabled==true( line 80 and line 97 ), in which case  you would not use some use supplied value you would use the actual service name of the devIdProvider.  So the conditionals here don't make any point either.

charts/chronicle/templates/tests/api-test.yaml
```diff
@@ -16,28 +16,62 @@ spec:
...
-              URL={{ include "chronicle.jwksUrl" . }}
-              scheme=$(echo $URL | cut -f 1 -d :)
-              hostAndPort=$(echo $URL | cut -f 3 -d /)
-              HOST=$(echo $hostAndPort | cut -f 1 -d :)
-              port=$(echo $hostAndPort | awk -F: '{print $2}')
-              if [ "$scheme" = "http" ]; then
-                defaultPort=80
-              elif [ "$scheme" = "https" ]; then
-                defaultPort=443
+              {{- if or (.Values.auth.jwks.url) (.Values.devIdProvider.enabled) }}
```
scealiontach, line 24: now you are doing this even in the case where auth.required==fale, and devIdProvider.enabled==false, in which case all of the following is going to go wrong.

charts/chronicle/values.yaml
```diff
@@ -41,6 +41,9 @@ extraVolumes: []
 ## @md | `extraVolumeMounts` | a list of additional volume mounts to add to chronicle | [] |
 extraVolumeMounts: []
 
+IdProvider:
+  service:
```
scealiontach, line 45: see above - I don't see the point of this value.

charts/chronicle/values.yaml
```diff
@@ -142,6 +144,11 @@ test:
       repository: blockchaintp/chronicle-helm-api-test-amd64
       ## @md | `test.api.image.tag` | the image tag | latest |
       tag: BTP2.1.0-0.7.3
+  auth:
+    ## @md | `test.auth.enabled` | true to enable testing with authorization header | true |
+    enabled: true
```
scealiontach, line 149: is there a time when you want this to be true, and yet have devIdProvider==false, no URLs provided, and auth.required==false?

scealiontach (changes requested):

charts/chronicle/templates/statefulset.yaml
```diff
@@ -103,8 +103,11 @@ spec:
           args:
             - |
               {{ if .Values.auth.required }}
+              {{ if and (not .Values.auth.jwks.url) (not .Values.devIdProvider.enabled) }}
```
scealiontach, line 106: better to invert this logic  from "A and !B" to "A or B" and switch up the if/else.  More readable.

scealiontach (changes requested):

scealiontach (approved):

mtbc (approved):

PR #346, opened 2023-06-12
reject request if --require-auth but missing id-claims

mtbc:
Fixes CHRON-395.

# PR Checklist


suchapalaver:
What's the relationship to whether or not the user provides `--id-claims` with this?

mtbc:
If they don't provide it, the default is `iss sub`. It'd interact in that, if a JWT is now failing due to `--require-auth`, they could choose different ID claims fields that *are* present in the JWT. However, choosing different fields has the effect of changing the user's Chronicle identity.

suchapalaver:
> If they don't provide it, the default is `iss sub`. It'd interact in that, if a JWT is now failing due to `--require-auth`, they could choose different ID claims fields that _are_ present in the JWT. However, choosing different fields has the effect of changing the user's Chronicle identity.

Thanks for confirming.

crates/api/src/chronicle_graphql/mod.rs
```diff
@@ -995,8 +996,13 @@ impl async_graphql::extensions::Extension for AuthFromJwt {
         next: async_graphql::extensions::NextPrepareRequest<'_>,
     ) -> async_graphql::ServerResult<async_graphql::Request> {
         if let Some(claims) = ctx.data_opt::<JwtClaims>() {
-            if let Some(chronicle_id) = self.identity(claims) {
-                request = request.data(chronicle_id);
+            match self.identity(claims) {
+                Ok(chronicle_id) => request = request.data(chronicle_id),
+                Err(error) if self.allow_anonymous => debug!("Identity check failed: {:?}", error),
```
suchapalaver, line 1001: This error message being same as on l.1003 and saying `failed` when it's not quite a failure in my book might be confusing in logs.

suchapalaver (approved):

crates/api/src/chronicle_graphql/mod.rs
mtbc: Thank you for the review! It seems rather "fail" to me because the user has authenticated and presented us with a token therefrom yet we cannot determine who they are. I'll add an extra note to this one, see if it then seems better.

suchapalaver (approved):

PR #358, opened 2023-06-15
fix(charts): fix auth endpoints cli templating

suchapalaver:
This PR corrects the templates to ensure the following behavior.

- If a user provides a jwks or userinfo endpoint (or both), Chronicle uses them
- If a user provides no endpoints but enables the `devIdProvider`, then Chronicle uses the endpoints for the `devIdProvider`
- If a user provides no endpoints and the `devIdProvider` is not enabled, we do not want any cli args for jwks and userinfo endpoints passed to Chronicle.

# PR Checklist

### DIFF
--- a/charts/chronicle/Chart.yaml
+++ b/charts/chronicle/Chart.yaml
@@ -17,7 +17,7 @@ keywords:
 # This is the chart version. This version number should be incremented each
 # time you make changes to the chart and its templates, including the app
 # version.
-version: 0.1.17
+version: 0.1.18
 
 # This is the version number of Chronicle being deployed. This version
 # number should be incremented each time you make changes to Chronicle.
--- a/charts/chronicle/templates/_chronicle.tpl
+++ b/charts/chronicle/templates/_chronicle.tpl
@@ -82,17 +82,20 @@ http://{{ include "chronicle.id-provider.service" . }}:8090/userinfo
 {{- else -}}
 {{- if .Values.devIdProvider.enabled -}}
 {{ include "chronicle.id-provider.service.jwks.url" . }}
-{{- else -}}
-{{/* Do nothing */}}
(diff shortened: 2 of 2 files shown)


mtbc:
It looks as if there's another case, which I'm okay with but may need clearly documenting: If the user supplies one of the endpoints and enables the dev-id-provider then the provider's used for the other endpoint that the user didn't supply. Basically: if you enable the dev-id-provider then both endpoints will be set, where user-provided URIs take priority over dev-id-provider ones.

suchapalaver:
> It looks as if there's another case, which I'm okay with but may need clearly documenting: If the user supplies one of the endpoints and enables the dev-id-provider then the provider's used for the other endpoint that the user didn't supply. Basically: if you enable the dev-id-provider then both endpoints will be set, where user-provided URIs take priority over dev-id-provider ones.

@mtbc That is not the expected behavior for this PR. I didn't intend to be. And following your comment, I went through the different options and templated them out to check that things are behaving as I describe them. To confirm, the expected behavior is: if either `auth.jwks.url` **or** `auth.userinfo.url` are provided, Chronicle will be passed the values supplied by the user and not those of the `devIdProvider`. Even if the `devIdProvider` is enabled. Chronicle uses the `devIdProvider` endpoints only when the user provides no values for `auth.jwks.url` **and/or** `auth.userinfo.url` and `devIdProvider` is enabled, in which case Chronicle uses both `devIdProvider` endpoints.

mtbc:
Ah yes, just finishing retesting and it's exactly matching, earlier I must have failed to save a `values.yaml` from the editor or not waited long enough for the `helm uninstall` or something and ended up deploying not what I thought. :+1:

scealiontach (approved):

PR #377, opened 2023-06-28
Needn't depend on specific Rust SDK revision

mtbc:
The sought commit has now been merged into the `main` branch.

# PR Checklist


scealiontach (changes requested):
Probably still best to use a specific commit reference even if it has merged to main.  Otherwise, if someone commits something broken to main, then we'll get the break here too.

PR #381, opened 2023-07-06
capture package repository on console

mtbc:
# PR Checklist


scealiontach (changes requested):
The reason you tee like this is so that you can see in the build log what is happening.

mtbc:
In that case, let's make it happen: now pushed a different fix!

scealiontach (approved):

PR #396, opened 2023-07-21
adjust submission protobuf

mtbc:
Followup to #392 to complete CHRON-442.

# PR Checklist

### DIFF
--- a/crates/chronicle-protocol/src/protos/submission.proto
+++ b/crates/chronicle-protocol/src/protos/submission.proto
@@ -18,7 +18,20 @@ message OpaPolicy {
 message Submission {
   string version = 1;
   uint64 span_id = 2;
-  repeated string body_old = 3 [deprecated = true];
-  string identity = 4;
-  string body = 5;
+  repeated string body_old = 3 [ deprecated = true ];
+  oneof identity_variant {
+    string identity_old = 4 [ deprecated = true ];
+    IdentityMessageV1 identity = 5;
+  }
+  oneof body_variant {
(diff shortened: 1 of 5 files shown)


crates/chronicle-protocol/src/protos/submission.proto
```diff
@@ -18,7 +18,18 @@ message OpaPolicy {
...
   uint64 span_id = 2;
-  repeated string body_old = 3 [deprecated = true];
-  string identity = 4;
-  string body = 5;
+  repeated string body_old = 3 [ deprecated = true ];
+  oneof identity_variant {
+    string identity_old = 4 [ deprecated = true ];
+    IdentityMessageV1 identity = 5;
+  }
+  BodyMessageV1 body = 6;
```
scealiontach, line 26: still put this behind a oneof.  It won't make a difference to the serialized message, but it will affect the generated code.  Better to bite that bullet now.

crates/chronicle-protocol/src/protos/submission.proto
```diff
@@ -18,7 +18,18 @@ message OpaPolicy {
...
+}
+
+message BodyMessageV1 {
+  string json = 1;
```
scealiontach, line 30: Don't name fields after types.  Particularly in this case.  just use `body`, or even better `payload`

mtbc:
Yet to add another `oneof`.

scealiontach (changes requested):

scealiontach (approved):

PR #402, opened 2023-07-25
WIP: health/ endpoint

suchapalaver:
# WIP: [CHRON-450](https://blockchaintp.atlassian.net/browse/CHRON-450)

- [x] Wraps our Prometheus metrics listener as another `poem` `Route::at` alongside other `--offer-endpoints` (currently `data` and `graphql`) so it listens alongside them on a `/metrics` path off `API_LISTEN_SOCKET` - takes care of [CHRON-450](https://blockchaintp.atlassian.net/browse/CHRON-450)
- [x] Adds `health/` endpoint that check current metrics from depth charge and returns `500`
- [ ] Improve response by including JSON with current round trip time
- [ ] Review the need to parse `String` from rendered metrics
- [ ] Figure out defaults - coupling of health endpoint with health metrics and the latter's configuration 
- [ ] Update Helm Chart

### Please complete this checklist after opening your PR

[CHRON-450]: https://blockchaintp.atlassian.net/browse/CHRON-450?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/crates/api/src/chronicle_graphql/mod.rs
+++ b/crates/api/src/chronicle_graphql/mod.rs
@@ -54,7 +54,7 @@ use tracing::{debug, error, instrument, warn};
 use url::Url;
 
 use self::authorization::TokenChecker;
-use crate::{chronicle_graphql::health::Metrics, ApiDispatch, ApiError, StoreError};
+use crate::{chronicle_graphql::health::Health, ApiDispatch, ApiError, StoreError};
 
 #[macro_use]
 pub mod activity;
@@ -527,14 +527,21 @@ impl SecurityConf {
 pub struct Endpoints {
     serve_graphql: bool,
     serve_data: bool,
+    serve_health: bool,
     serve_metrics: bool,
 }
 
 impl Endpoints {
-    pub fn new(serve_graphql: bool, serve_data: bool, serve_metrics: bool) -> Self {
+    pub fn new(
+        serve_graphql: bool,
+        serve_data: bool,
+        serve_health: bool,
+        serve_metrics: bool,
+    ) -> Self {
         Self {
             serve_graphql,
             serve_data,
+            serve_health,
             serve_metrics,
         }
     }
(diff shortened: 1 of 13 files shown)


crates/api/src/chronicle_graphql/mod.rs
```diff
@@ -522,6 +524,29 @@ impl SecurityConf {
     }
 }
 
+pub struct Endpoints {
```
mtbc, line 527: Perhaps a set of enum values?

mtbc: (That's just me thinking aloud, it's fine if you prefer it as it is, feel free to resolve this.)

ryanbtp: They are not all mutually exclusive or related, but I agree seeing more than 2 bools at once is a worry. We may be pushing complexity up to configuration here. Simple data / graphql are always reasonable to assume. As they have known base urls, infrastructure can be used to exclude.

It's only health / metrics thats a choice here. Our current requirement is health, we can deal with exposing a prometheus endpoint properly later.

This is my fault, as I thought a prometheus endpoint would be basically free if we based health on metrics - apparently not the case but we will be able to figure it out.

ryanbtp (changes requested):

crates/api/src/chronicle_graphql/mod.rs
mtbc: Though, I'm happy for none of this to be figured out in this PR if we'd prefer a separate ticket for endpoint conf (or not), at least it's just simple on-off configuration for separable things.

suchapalaver: This was simply because the compiler was complaining about too many args in a function so I put them in a struct together.

PR #404, opened 2023-07-29
WIP/Experiment: perftest

suchapalaver:
# WIP: Performance test

## [CHRON-467](https://blockchaintp.atlassian.net/browse/CHRON-467)

- adds a `perftest` sub-command that takes an argument for the number of Chronicle commands to be dispatched and timed, and a `cap` argument for the number of submissions allowed to run per second. The task dispatches `x` number of Chronicle commands, writing them to the `system` namespace, and then waits for `x` number of confirmations of commits or failures before recording the elapsed time. 
- See [results](https://blockchaintp.atlassian.net/l/cp/1DkgDatf)

## Updates 

More recent updates [here](https://blockchaintp.atlassian.net/l/cp/1DkgDatf)

### July 31
Altering the `chronicle` container Chart `args` to `tail -f /dev/null` and `exec`ing into the container to run the `perftest` command. To @ryanbtp's point in the discussion, I've found that running 50 transactions from the perftest can be done repeatedly, manually rerunning the command after completion gives us a set of the same results. Upping the number of commands to 100 shows the exponential pattern we see elsewhere. Will explore introducing pauses to the `perftest` code.

## issues: 
- @ryanbtp, @mtbc, different approaches to tracking commit messages and `tx_id`s affects results; 
    - I tested this with renditions that ensured the commit messages matched to guarantee we can rely on this being the case, i.e. send forty `perftest` commands and receiving forty commit messages means all forty transactions were successful. Reverting back to not checking because of the cost of doing so.
- @scealiontach, still trying to nail down requirements regarding test conditions and what's "good enough" - is installing the chart and running the performance test a decent measure, or do we need to be able to keep running multiple tests after we have already written to Sawtooth in a previous test? [Multiple runs](https://blockchaintp.atlassian.net/l/cp/o0QT8mGF) are showing an exponential growth in elapsed times. 
    - I've stopped installing a `perftest` configured `chronicle` Helm Chart and started using `port-forward` instead. Getting slow - ~ 40 seconds for 10 commands - results. @ryanbtp, @mtbc, I'm going to investigate further what this means in terms of @ryanbtp's points below about the `Api` initialization and getting the last commit block.   

## installing Chart configured to `perftest`
**TL;DR** - restart issues because the Chronicle `Pod` is coupled to the postgres database, exponential growth in times taken with each restart until it eventually just chokes up.

- It does restart itself, but sometimes needs a (`kubectl delete pod perftest-chronicle-chronicle-0`) kick due to `Lagged` receiver errors after the container restarts.
- Using the Helm configuration here means that the Chronicle `Statefulset` will start up in `perftest` mode and begin repeatedly running the performance test. Once finished the process will simply restart. Logs will need to be observed for now as they will disappear once the Pod restarts after each performance test is done. 
- For results see the [Chronicle perftest page](https://blockchaintp.atlassian.net/l/cp/o0QT8mGF) in Atlassian.

**Note** that first runs seem to contain a number of queuing errors that are not present second or third time around. But let's see. I've been running 2500 mutations each round in the `perftest` namespace on `azcue-1`. Things really start to slow down and then grind to a halt after a few rounds. And now trying 1000. Note also, for ease of looking at the results, I've used `info` level for logs related to the performance testing. `kubectl logs --namespace=perftest perftest-chronicle-chronicle-0 -c chronicle -f` to view the relevant logs. After about 5 runs through we start seeing a **lot** more queuing errors. 

Look out for these messages for an indication of how many commands are left before the test ends its current round:

```json
{"timestamp":"2023-07-29T14:18:08.036048Z","level":"INFO","fields":{"message":"1042 operations remaining"},"target":"chronicle::bootstrap","threadId":"ThreadId(7)"}
```

Look out for the following at the end of the logs when the task has completed:

Perftest complete:

```json
{"timestamp":"2023-07-29T14:32:19.613629Z","level":"INFO","fields":{"message":"Perftest complete"},"target":"chronicle::bootstrap","span":{"config":"Config { secrets: SecretConfig { path: \"/var/lib/chronicle/secrets/\" }, validator: ValidatorConfig { address: Url { scheme: \"tcp\", cannot_be_a_base: false, username: \"\", password: ***SECRET***, host: Some(Domain(\"perftest-sawtooth.perftest.svc.cluster.local\")), port: Some(4004), path: \"\", query: None, fragment: None } }, namespace_bindings: {\"default\": fd717fd6-70f1-44c1-81de-287d5e101089} }","name":"execute_subcommand"},"spans":[{"config":"Config { secrets: SecretConfig { path: \"/var/lib/chronicle/secrets/\" }, validator: ValidatorConfig { address: Url { scheme: \"tcp\", cannot_be_a_base: false, username: \"\", password: ***SECRET***, host: Some(Domain(\"perftest-sawtooth.perftest.svc.cluster.local\")), port: Some(4004), path: \"\", query: None, fragment: None } }, namespace_bindings: {\"default\": fd717fd6-70f1-44c1-81de-287d5e101089} }","name":"execute_subcommand"}],"threadId":"ThreadId(1)"}
```

How many seconds the test took to reach completion:

```json
{"timestamp":"2023-07-29T14:32:19.613676Z","level":"INFO","fields":{"message":"Perftest took 1643 seconds"},"target":"chronicle::bootstrap","span":{"config":"Config { secrets: SecretConfig { path: \"/var/lib/chronicle/secrets/\" }, validator: ValidatorConfig { address: Url { scheme: \"tcp\", cannot_be_a_base: false, username: \"\", password: ***SECRET***, host: Some(Domain(\"perftest-sawtooth.perftest.svc.cluster.local\")), port: Some(4004), path: \"\", query: None, fragment: None } }, namespace_bindings: {\"default\": fd717fd6-70f1-44c1-81de-287d5e101089} }","name":"execute_subcommand"},"spans":[{"config":"Config { secrets: SecretConfig { path: \"/var/lib/chronicle/secrets/\" }, validator: ValidatorConfig { address: Url { scheme: \"tcp\", cannot_be_a_base: false, username: \"\", password: ***SECRET***, host: Some(Domain(\"perftest-sawtooth.perftest.svc.cluster.local\")), port: Some(4004), path: \"\", query: None, fragment: None } }, namespace_bindings: {\"default\": fd717fd6-70f1-44c1-81de-287d5e101089} }","name":"execute_subcommand"}],"threadId":"ThreadId(1)"}
```

How many operations failed:

```json
{"timestamp":"2023-07-29T14:32:19.613697Z","level":"INFO","fields":{"message":"0 operations failed"},"target":"chronicle::bootstrap","span":{"config":"Config { secrets: SecretConfig { path: \"/var/lib/chronicle/secrets/\" }, validator: ValidatorConfig { address: Url { scheme: \"tcp\", cannot_be_a_base: false, username: \"\", password: ***SECRET***, host: Some(Domain(\"perftest-sawtooth.perftest.svc.cluster.local\")), port: Some(4004), path: \"\", query: None, fragment: None } }, namespace_bindings: {\"default\": fd717fd6-70f1-44c1-81de-287d5e101089} }","name":"execute_subcommand"},"spans":[{"config":"Config { secrets: SecretConfig { path: \"/var/lib/chronicle/secrets/\" }, validator: ValidatorConfig { address: Url { scheme: \"tcp\", cannot_be_a_base: false, username: \"\", password: ***SECRET***, host: Some(Domain(\"perftest-sawtooth.perftest.svc.cluster.local\")), port: Some(4004), path: \"\", query: None, fragment: None } }, namespace_bindings: {\"default\": fd717fd6-70f1-44c1-81de-287d5e101089} }","name":"execute_subcommand"}],"threadId":"ThreadId(1)"}
```

- This can be configured with a Helm `values.yaml` file like the one below. Note that we are using images built from source:

```yaml
---
perftest:
  enabled: true
  ops: 2500

imagePullSecrets:
  enabled: true
  value: 
    - name: regcred
    
image: 
  pullPolicy: Always
  repository: dev.catenasys.com:8083/joseph/perftest
  tag: latest
  
logLevel: info,cranelift_codegen=off,wasmtime_cranelift=off,regalloc2::ion=off,cranelift_wasm=off
  
opa:
  enabled: true
  opaInit:
    image:
      pullPolicy: Always
      repository: dev.catenasys.com:8083/joseph/opactl-amd64
      tag: latest
  tp:
    image:
      repository: dev.catenasys.com:8083/joseph/opa-tp-amd64
      tag: latest
      pullPolicy: Always

tp:
  image:
    pullPolicy: Always
    repository: dev.catenasys.com:8083/joseph/chronicle-tp-amd64
    tag: latest    
```

[CHRON-467]: https://blockchaintp.atlassian.net/browse/CHRON-467?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/crates/api/src/lib.rs
+++ b/crates/api/src/lib.rs
@@ -1943,7 +1943,7 @@ mod test {
     }
 
     fn embed_chronicle_tp() -> EmbeddedChronicleTp {
-        chronicle_telemetry::telemetry(None, chronicle_telemetry::ConsoleLogging::Pretty);
+        chronicle_telemetry::telemetry(None, None, chronicle_telemetry::ConsoleLogging::Pretty);
         let mut buf = vec![];
         Setting {
             entries: vec![Setting_Entry {
@@ -2004,7 +2004,7 @@ mod test {
     }
 
     async fn test_api<'a>() -> TestDispatch<'a> {
-        chronicle_telemetry::telemetry(None, chronicle_telemetry::ConsoleLogging::Pretty);
+        chronicle_telemetry::telemetry(None, None, chronicle_telemetry::ConsoleLogging::Pretty);
 
         let secretpath = TempDir::new().unwrap().into_path();
 
--- a/crates/api/src/lib.rs
+++ b/crates/api/src/lib.rs
@@ -19,7 +19,7 @@ use diesel_migrations::MigrationHarness;
 use futures::{select, FutureExt, StreamExt};
 
 use common::{
-    attributes::Attributes,
+    attributes::{Attribute, Attributes},
     commands::*,
     identity::{AuthId, IdentityError},
     k256::ecdsa::SigningKey,
@@ -33,8 +33,8 @@ use common::{
         },
         to_json_ld::ToJson,
         ActivityId, AgentId, ChronicleIri, ChronicleTransaction, ChronicleTransactionId,
-        Contradiction, EntityId, ExternalId, ExternalIdPart, NamespaceId, ProcessorError,
-        ProvModel, Role, SYSTEM_ID, SYSTEM_UUID,
+        Contradiction, DomaintypeId, EntityId, ExternalId, ExternalIdPart, NamespaceId,
+        ProcessorError, ProvModel, Role, SYSTEM_ID, SYSTEM_UUID,
     },
     signing::{DirectoryStoredKeys, SignerError},
 };
@@ -55,7 +55,10 @@ use std::{
 };
 use thiserror::Error;
 use tokio::{
-    sync::mpsc::{self, error::SendError, Sender},
+    sync::{
+        broadcast::error::RecvError,
+        mpsc::{self, error::SendError, Sender},
+    },
     task::JoinError,
 };
 
@@ -67,6 +70,9 @@ use uuid::Uuid;
 
 #[derive(Error, Debug)]
 pub enum ApiError {
+    #[error("Failure in commit notification stream: {0}")]
+    CommitNoticiationStream(#[from] RecvError),
+
     #[error("Storage: {0:?}")]
     Store(#[from] persistence::StoreError),
 
@@ -129,6 +135,9 @@ pub enum ApiError {
 
     #[error("Authentication endpoint error: {0}")]
     AuthenticationEndpoint(#[from] chronicle_graphql::AuthorizationError),
+
+    #[error("Perftest unable to complete")]
+    PerftestError,
 }
 
 /// Ugly but we need this until ! is stable, see <https://github.com/rust-lang/rust/issues/64715>
@@ -252,6 +261,43 @@ impl ApiDispatch {
         self.dispatch(
             ApiCommand::DepthCharge(DepthChargeCommand { namespace }),
             identity.clone(),
+        )
+        .await
+    }
+
+    #[instrument]
+    pub async fn handle_perftest(
+        &self,
+        id: AuthId,
+        namespace: NamespaceId,
+    ) -> Result<ApiResponse, ApiError> {
+        self.dispatch_perftest(id, namespace).await
+    }
+
+    #[instrument]
+    async fn dispatch_perftest(
+        &self,
+        identity: AuthId,
+        namespace: NamespaceId,
+    ) -> Result<ApiResponse, ApiError> {
+        self.dispatch(
+            ApiCommand::Activity(ActivityCommand::Create {
+                external_id: Uuid::new_v4().to_string().into(),
+                namespace: common::prov::SYSTEM_ID.into(),
+                attributes: Attributes {
+                    typ: Some(DomaintypeId::from_external_id("perftest")),
+                    attributes: [(
+                        "perftest".to_owned(),
+                        Attribute {
+                            typ: "perftest".to_owned(),
+                            value: serde_json::Value::String("perftest".to_owned()),
+                        },
+                    )]
+                    .into_iter()
+                    .collect(),
+                },
+            }),
+            identity,
         )
         .await
     }
--- a/crates/api/src/lib.rs
+++ b/crates/api/src/lib.rs
@@ -27,7 +27,7 @@ use common::{
     prov::{
         operations::{
             ActivityExists, ActivityUses, ActsOnBehalfOf, AgentExists, ChronicleOperation,
-            CreateNamespace, DerivationType, EndActivity, EntityDerive, EntityExists,
+            CreateNamespace, DerivationType, EndActivity, EntityDerive, EntityExists, RegisterKey,
             SetAttributes, StartActivity, WasAssociatedWith, WasAttributedTo, WasGeneratedBy,
             WasInformedBy,
         },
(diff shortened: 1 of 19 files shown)


crates/chronicle/src/bootstrap/mod.rs
```diff
@@ -489,6 +496,66 @@ where
...
+            info!("Perftest operations sent");
+        });
+
+        let perftest_notify_api = api.clone();
+
+        let (elapsed_time, failed_submissions) = tokio::task::spawn(async move {
+            let mut failed_submissions = 0;
+            let mut tx_notifications = perftest_notify_api.notify_commit.subscribe();
+            while ops_copy > 0 {
+                let stage = tx_notifications.recv().await?;
```
suchapalaver, line 524: If I move this two lines above, outside the `while` loop, performing 2500 transactions completes in 282 seconds, but with 1607 rejections due to queuing errors.

crates/chronicle/src/bootstrap/mod.rs
```diff
@@ -489,6 +496,66 @@ where
...
+            }
+            info!("Perftest operations sent");
+        });
+
+        let perftest_notify_api = api.clone();
+
+        let (elapsed_time, failed_submissions) = tokio::task::spawn(async move {
+            let mut failed_submissions = 0;
+            while ops_copy > 0 {
+                let mut tx_notifications = perftest_notify_api.notify_commit.subscribe();
```
suchapalaver, line 523: If I move this two lines above, outside the while loop, performing 2500 transactions completes in 282 seconds, but with 1607 rejections due to queuing errors.

crates/chronicle/src/bootstrap/mod.rs
```diff
@@ -489,6 +493,33 @@ where
...
+        tokio::task::spawn(async move {
+            while ops > 0 {
+                let identity = AuthId::chronicle();
+                let namespace = system_namespace();
+                let _res = perftest_ops_api.handle_perftest(identity, namespace).await;
+                ops -= 1;
+            }
+        });
+
+        let mut tx_notifications = api.notify_commit.subscribe();
```
mtbc, line 511: Should we subscribe _before_ we spawn the other task?

ryanbtp: It will subscribe from the stored block height:

```
       let (from_block_id, from_position) = {
            match &from_block_id {
                FromBlock::First => {
                    let first_block = self.get_first_block().await?;
                    debug!(first_block = %first_block);
                    (first_block, None)
                }
                FromBlock::Head => {
                    let (block_num, block_id) = self.get_block_height().await?;
                    debug!(block_height = %block_num, block_id = %block_id);
                    (block_id, Some(block_num))
                }
                FromBlock::BlockId(block_id) => (*block_id, None),
            }
        };
        ```

**Comment by ryanbtp** [comment]
What would be useful here is to be able to set (or at least cap) operations per second, rather than just the total number. We are obviously saturating it here, but we need to be able to find the stable point. Usually with performance / failure there's a section where there's a near linear relationship between load and latency, then when things start to go off there will be a geometric / exponential that then starts producing errors

File: crates/chronicle/src/bootstrap/mod.rs (hunk 3 above)
Comment (ryanbtp) [reply crates/chronicle/src/bootstrap/mod.rs:511]: And this may explain worse behaviour with restarts under load

**Comment by ryanbtp** [comment]
Lag errors are interesting, as it implies chronicle cannot write to the store quick enough - quite possible with heavy loads. We can increase queue lengths to smooth things out once we know what sort of numbers are sane

File: crates/chronicle/src/bootstrap/mod.rs (hunk 3 above)
Comment (mtbc) [reply crates/chronicle/src/bootstrap/mod.rs]: Ah, the `last_seen_block` in `Api::new`? :+1:

File: crates/api/src/lib.rs [hunk 4]
```diff
@@ -67,6 +70,9 @@ use uuid::Uuid;
 
 #[derive(Error, Debug)]
 pub enum ApiError {
+    #[error("Failure in commit notification stream: {0}")]
+    CommitNoticiationStream(#[from] RecvError),
```
mtbc, line 74:
```suggestion
    CommitNotificationStream(#[from] RecvError),
```

PR #213, opened 2023-03-02
API calls resulting in no data will not be dispatched

suchapalaver:
Adds API methods and responses checking whether Chronicle operations resulting from API calls will result in changes in state. If they don't, they won't be submitted to the Transaction Processor and return `Submission::AlreadyRecorded` along with GraphQL data containing `context` and a `null` `tx_id`.

---
[CHRON-146](https://blockchaintp.atlassian.net/browse/CHRON-146)

Signed-off-by: Joseph Livesey [joseph.livesey@btp.works](mailto:joseph.livesey@btp.works)

[CHRON-146]: https://blockchaintp.atlassian.net/browse/CHRON-146?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #408, opened 2023-08-04
fix(api): streamline checks that api calls have data effects

suchapalaver:
# [CHRON-471](https://blockchaintp.atlassian.net/browse/CHRON-471)

Whereas we [were](https://github.com/btpworks/chronicle/pull/213) querying the entire namespace in question when checking whether applying an operation would have effects, we now query much smaller pieces of data to accomplish the same checks. 

Can confirm checking `for i in {1..100}; do cargo run --bin chronicle --features inmem -- perftest 50 60; done
` and the result remained 1 second each time. Prior to these changes we were seeing exponential growth in the elapsed times.

### Please complete this checklist after opening your PR

[CHRON-471]: https://blockchaintp.atlassian.net/browse/CHRON-471?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/crates/api/src/lib.rs
+++ b/crates/api/src/lib.rs
@@ -27,7 +27,7 @@ use common::{
     prov::{
         operations::{
             ActivityExists, ActivityUses, ActsOnBehalfOf, AgentExists, ChronicleOperation,
-            CreateNamespace, DerivationType, EndActivity, EntityDerive, EntityExists,
+            CreateNamespace, DerivationType, EndActivity, EntityDerive, EntityExists, RegisterKey,
             SetAttributes, StartActivity, WasAssociatedWith, WasAttributedTo, WasGeneratedBy,
             WasInformedBy,
         },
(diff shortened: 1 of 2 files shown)


crates/api/src/lib.rs
```diff
@@ -563,31 +563,291 @@ where
...
+                        model,
+                        agent_id,
+                        namespace.external_id_part(),
+                    )?
+                }
+            };
+            let state = applied_model.clone();
+            applied_model.apply(op)?;
+            if state != applied_model {
+                info!("Compare: {:#?} != {:#?}", state, applied_model);
```
mtbc, line 842: Probably push this up to debug or trace?

suchapalaver: Leftovers from debugging. Thanks

mtbc:
This looks plausible to me and it works well and I am really glad we have this fix but I would appreciate a second pair of eyes as I could easily be missing something.

ryanbtp (approved):

PR #416, opened 2023-10-11
refactor code to pass cargo clippy tests

mtbc:
Should allow passing the GitHub workflows.

# PR Checklist

## Please complete this checklist after opening your PR

* [x] I have updated documentation as necessary
* [x] I have updated helm chart(s) if needed
* [x] I have added tests if needed
* [ ] All new and existing tests are passing
* [x] Any breaking changes are clearly flagged and documented
* [x] I’ve included a link to any relevant ticket(s)

PR #418, opened 2023-10-23
updates to documentation and insertion of cookbooks

ChrisH1066:
# PR Checklist


mtbc:
#416 should fix the `cargo-checkmate` issues. The DCO failure is that you need the `Signed-off-by:` bit in your commit message.

docs/assets/logos/favicon.png
duncanjw, line 1: Presumably we should change this and use the Chronicle marque

duncanjw (changes requested):
Installing with Sextant should be replaced with Installing with Rancher and either the cookbook stuff pulled into here or at the very least referenced. I prefer the former. Either way we will need to get SUSE to update any cookbook links etc

ryan-s-roberts (approved):

